PR #18, opened 2026-04-03
feat: custom archive format (-F custom) (#13)

### LINKED ISSUE #13: feat: custom archive format (-F custom)
## Goal
Implement `pg_dump -F custom` (compressed binary archive format), prerequisite for parallel restore.

## Tests to un-ignore (t002_pg_dump.rs)
- `run_defaults_custom_format`
- Any restore tests that use custom format input

## What to implement
In `src/dump/format.rs` and `src/restore/`:
- Write pg_dump custom archive format (.dump files)
- Custom format = zlib-compressed, object-by-object, with TOC header
- Reference: https://github.com/postgres/postgres/blob/master/src/bin/pg_dump/pg_backup_archiver.c
- pg_restore must be able to read custom format produced by our tool

## Definition of done
- Target tests GREEN
- Round-trip: `pg_dump -F custom` → `pg_restore` restores correctly
- cargo test passes, clippy clean, fmt clean
- PR created targeting main

NikolayS:
Closes #13

Implements pg_dump custom archive format (binary, zlib-compressed):
- PGDMP header + TOC entries + compressed data blocks
- pg_restore can read and restore custom format dumps
- Round-trip test: dump -F custom → restore → verify row count
- Added flate2 dependency for zlib compression

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 4
 
 [[package]]
+name = "adler2"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
+
+[[package]]
 name = "anstream"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -193,6 +199,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "crc32fast"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9481c1c90cbf2ac953f07c8d4a58aa3945c425b7185c9154d67a65e4230da511"
+dependencies = [
+ "cfg-if",
+]
+
+[[package]]
 name = "crypto-common"
 version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -249,6 +264,16 @@ name = "fastrand"
 version = "2.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37909eebbb50d72f9059c3b6d82c0463f2ff062c9e95845c43a6c9c0355411be"
+
+[[package]]
+name = "flate2"
+version = "1.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "843fba2746e448b37e26a819579957415c8cef339bf08564fe8b7ddbd959573c"
+dependencies = [
+ "crc32fast",
+ "miniz_oxide",
+]
 
 [[package]]
 name = "foldhash"
@@ -461,6 +486,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f8ca58f447f06ed17d5fc4043ce1b10dd205e060fb3ce5b979b8ed8e59ff3f79"
 
 [[package]]
+name = "miniz_oxide"
+version = "0.8.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1fa76a2c86f704bdb222d66965fb3d63269ce38518b83cb0575fca855ebb6316"
+dependencies = [
+ "adler2",
+ "simd-adler32",
+]
+
+[[package]]
 name = "mio"
 version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -537,6 +572,7 @@ dependencies = [
  "anyhow",
  "bytes",
  "clap",
+ "flate2",
  "futures-util",
  "tempfile",
  "thiserror",
@@ -751,6 +787,12 @@ dependencies = [
  "errno",
  "libc",
 ]
+
+[[package]]
+name = "simd-adler32"
+version = "0.3.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "703d5c7ef118737c72f1af64ad2f6f8c5e1921f818cdcb97b8fe6fc69bf66214"
 
 [[package]]
 name = "siphasher"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ license = "MIT"
 anyhow = "1"
 bytes = "1"
 clap = { version = "4", features = ["derive"] }
+flate2 = "1"
 futures-util = "0.3"
 thiserror = "2"
 tokio = { version = "1", features = ["full"] }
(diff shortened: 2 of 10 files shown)


NikolayS:
## REV Code Review — PR #18: custom archive format (-F custom)

**Author:** @NikolayS
**CI:** ✅ Build & Test passed

---

### BLOCKING ISSUES (1)

#### 🔴 HIGH — `read_str` cannot distinguish NULL from empty string (confidence: 9)

**`src/dump/custom_format.rs`, `read_toc_entry`**

`write_null_str` writes `-1` as the dependency-list terminator, but `read_str` converts both `len == -1` (NULL) and `len == 0` (empty) to `""`. The loop in `read_toc_entry` breaks on `""`:

```rust
// writer:
write_null_str(w)?;  // writes -1i32.to_le_bytes()

// reader:
pub fn read_str(r: &mut impl std::io::Read) -> io::Result<String> {
    let len = i32::from_le_bytes(len_bytes);
    if len < 0 {
        return Ok(String::new());  // NULL → ""
    }
    // ...
}
// loop breaks on "" — can't distinguish NULL terminator from
// a genuine zero-length string dependency ID
```

This works today because dependency IDs are always non-empty integers, but any table entry with a zero-length string in that position would silently truncate the dependency list, potentially losing dependencies and causing incorrect restore ordering.

**Fix:** Add a distinct `read_null_str` / change `read_str` to return `Option<String>` where `None` means NULL, or change the terminator to a sentinel integer (e.g., write `write_int(w, -1)` as the terminator) and parse accordingly. Alternatively: check the raw length bytes before deciding to break vs. continue.

---

### NON-BLOCKING (2)

#### 🟡 MEDIUM — `order by 1` in `write_table_data_to_string` (confidence: 8)

**`src/dump/format.rs`, line ~55**

```rust
let query = format!("select {col_list} from {qname} order by 1");
```

`ORDER BY 1` sorts by the first column only. For tables where the first column is nullable, non-unique, or of a non-orderable type (e.g., `json`, `array`), this either produces non-deterministic output or fails at runtime. pg_dump uses no ordering by default (relies on heap order) or sorts by PK. This SQL style also violates postgres-ai rules (explicit column names in ORDER BY).

**Fix:** Either remove the `ORDER BY` entirely (consistent with real pg_dump behavior) or use `ORDER BY` with explicit primary key columns if determinism is needed.

#### 🟡 INFO — Dead code: `data_toc_start` computed but unused (confidence: 8)

**`src/dump/mod.rs`, `dump_custom`**

```rust
let data_toc_start = toc_buf.len();
// ...
let _ = data_toc_start;  // suppressed immediately
```

The variable was likely an intermediate step during development. `data_entry_offsets.first()` is used for patching instead. This is dead code that adds confusion.

**Fix:** Remove `data_toc_start` and the `let _ = data_toc_start;` suppressor.

---

### POTENTIAL ISSUES (4)

#### 🟠 MEDIUM — No bounds check before `vec![0u8; compressed_size as usize]` (confidence: 6)

**`src/dump/custom_format.rs`, `read_next_data_block`**

```rust
let compressed_size = read_int(r)?;
// ...
let mut compressed = vec![0u8; compressed_size as usize];
```

If `compressed_size` is a very large positive value (e.g., from a malformed/corrupt archive), this will attempt to allocate gigabytes of memory. There's a `<= 0` guard but no upper bound.

**Fix:** Add a sanity cap, e.g. `if compressed_size > 256 * 1024 * 1024 { return Err(...) }`.

#### 🟠 MEDIUM — Missing test: empty table in custom format (confidence: 6)

**`tests/pg_dump/t002_pg_dump.rs`**

The round-trip test only exercises 3-row happy path. An empty table produces `data_str = ""` → `data.is_empty() in Some((_dump_id, data)) if data.is_empty()` guard skips the COPY. But there's no test confirming that an empty-table restore doesn't crash or corrupt the restore state machine.

**Fix:** Add a test variant with a table that has 0 rows.

#### 🟠 LOW — Missing test: corrupt / truncated archive input (confidence: 5)

**`src/restore/mod.rs`, `restore_custom`**

No test verifies that `restore_custom` returns a proper error (not a panic) on truncated or corrupt input bytes.

**Fix:** Add a unit test: `restore_custom(b"PGDMP\x01\x0f\x00\x04\x08\x03<truncated>", &opts)` → should return `Err(...)`.

#### 🟠 INFO — `Copyright 2026` in new file (confidence: 5)

**`src/dump/custom_format.rs`, line 1**

```rust
// Copyright 2026 pg_plumbing contributors
```

Year 2026 is likely a typo — should be 2025 (or whichever is current).

**Fix:** Correct the copyright year if this is not intentional.

---

### Summary table

| Area | Blocking | Non-Blocking | Potential |
|------|----------|--------------|-----------|
| Security | 0 | 0 | 0 |
| Bugs | 1 | 1 | 1 |
| Tests | 0 | 0 | 2 |
| Guidelines | 0 | 1 | 1 |
| **Total** | **1** | **2** | **4** |

---

NikolayS:
## REV Re-review (post-fix)

**CI:** ✅ Build & Test passed on latest commit (`12cb97f5`)

---

### Previous blocking issue: Fixed ✅

`read_opt_str()` is correct. It reads a 4-byte LE length, returns `None` for `len < 0` (the NULL sentinel written by `write_null_str`) and `Some(s)` for `len >= 0` — correctly distinguishing NULL from empty strings. The dependency-list loop in `read_toc_entry` now breaks only on the NULL terminator, not on any empty string dep ID. The encoding is consistent with the writer: `write_null_str` emits `-1i32` as 4 bytes LE, and `read_opt_str` reads exactly 4 bytes. ✓

### Previous non-blocking issues: Both fixed ✅

- **`ORDER BY 1` removed** from `write_table_data` — the `format!(...from {qname})` query is now order-agnostic. The new `write_table_data_to_string` (for custom format) never had the ORDER BY in the first place. ✓
- **Dead code removed** — `eprintln!("not yet implemented")` stubs in both binaries replaced with real implementation; empty `#[ignore]` test stub filled in with the full round-trip test. ✓

---

### New issues found

**Non-blocking (minor):** `write_data_block` casts `compressed.len() as i32` when writing the block size field. For tables with > ~2 GB of compressed data this would overflow (silent wraparound in release mode, panic in debug). Not a concern for the current scope, but worth a `u32` cast + overflow check before writing to the size field if large-table support is planned.

**Non-blocking (minor):** In `read_next_data_block`, if the end-of-data marker byte is not `BLK_DATA` (i.e. a malformed/truncated archive), the byte is silently consumed and the function returns the decompressed data as-is. The next call would then misparse the stream. A small defensive check (`return Err(...)` if `marker_type[0] != BLK_DATA`) would make error reporting cleaner, but this is an edge-case on corrupted input.

No new blocking issues introduced.

---

**Verdict:** APPROVED ✅ — no blocking issues. The core sentinel fix is correct, both prior non-blocking issues are resolved cleanly, and CI is green.

---
*REV-assisted re-review*

Issue #1, opened 2026-04-03
feat: scaffold Cargo project with CI workflow and basic CLI


NikolayS:
## Sprint 1 — Task 2: Scaffold

Set up the foundational project structure so all other tasks have something to build on.

### Deliverables

- `Cargo.toml` workspace (binary crate `pg_plumbing`, lib crate)
- `src/main.rs` — `pg_dump` / `pg_restore` subcommands via `clap`
- `src/lib.rs` — shared library stub
- `src/dump/mod.rs`, `src/restore/mod.rs` — module stubs
- `.github/workflows/ci.yml` — runs `cargo test` + `cargo clippy` on PG 14-18 (use services: postgres matrix)
- `tests/` directory structure as specified in SPEC.md

### Acceptance criteria

- `cargo build` succeeds with zero warnings
- `cargo clippy -- -D warnings` passes
- `cargo test` runs (0 tests passing is fine at this stage)
- CI workflow file is valid YAML and triggers on push/PR
- Binary exposes `pg_dump` and `pg_restore` subcommands (or separate binaries)

### Notes

- Follow CLAUDE.md style rules
- Use `anyhow` for CLI errors, `thiserror` for library errors
- Use `clap` with derive macros
- Conventional commit: `feat: scaffold cargo project and CI workflow`

NikolayS:
Update on libpgdump: created 4 days ago, 1 star, 0 forks. Too young to depend on. Use as **reference only** for understanding the pg_dump archive format — implement our own parser for full control over streaming, parallelism, and progress.

PR #19, opened 2026-04-03
feat: directory format dump/restore (-F directory) (#14)

NikolayS:
Closes #14

Implements pg_dump directory format:
- Creates output directory with toc.dat + per-table .dat files
- pg_restore detects directory input, reads toc.dat + data files
- Round-trip test: dump -F directory → restore → verify row count

### DIFF
--- a/src/dump/directory_format.rs
+++ b/src/dump/directory_format.rs
@@ -1,0 +1,182 @@
+// Copyright 2026 pg_plumbing contributors
+// SPDX-License-Identifier: MIT
+
+//! Directory format dump (`pg_dump -F directory`).
+//!
+//! Creates a directory containing:
+//!   - `toc.dat` — text TOC listing schema DDL and data file references
+//!   - `<N>.dat`  — one plain COPY-format file per table with data
+
+use anyhow::{bail, Context, Result};
+use tokio_postgres::{Client, NoTls};
+
+use super::catalog;
+use super::format;
+use super::DumpOptions;
+
+/// Dump a database in directory format.
+///
+/// Creates `output_dir` (errors if it already exists and is non-empty),
+/// writes `toc.dat` and one `<N>.dat` per table.
+pub async fn dump_directory(opts: &DumpOptions, output_dir: &str) -> Result<()> {
+    // Create the output directory; fail if it already exists and is non-empty.
+    let dir_path = std::path::Path::new(output_dir);
+    if dir_path.exists() {
+        let entries: Vec<_> = std::fs::read_dir(dir_path)
+            .context("failed to read output directory")?
+            .collect();
+        if !entries.is_empty() {
+            bail!("directory \"{}\" exists and is not empty", output_dir);
+        }
+    } else {
+        std::fs::create_dir_all(dir_path)
+            .with_context(|| format!("failed to create directory \"{}\"", output_dir))?;
+    }
+
+    // Connect to the database.
+    let conninfo = crate::build_conninfo(&opts.dbname);
+    let (client, connection) = tokio_postgres::connect(&conninfo, NoTls)
+        .await
+        .with_context(|| format!("failed to connect to database \"{}\"", opts.dbname))?;
+
+    tokio::spawn(async move {
+        if let Err(e) = connection.await {
+            eprintln!("connection error: {e}");
+        }
+    });
+
+    let tables = catalog::get_tables(&client, opts)
+        .await
+        .context("failed to query catalog")?;
+
+    // Build the TOC and per-table data files.
+    let mut toc = String::new();
+
+    toc.push_str("; pg_plumbing directory format TOC\n");
+    toc.push_str(&format!("; dbname: {}\n", opts.dbname));
+    toc.push_str(";\n");
+
+    // Schema section.
+    if !opts.data_only {
+        for table in &tables {
+            // Dump any sequences that this table's columns depend on first.
+            let sequences = get_table_sequences(&client, table).await?;
+            for (seq_schema, seq_name, seq_ddl) in &sequences {
+                let seq_key = format!("{seq_schema}.{seq_name}");
+                let seq_file = format!("{seq_name}.seq.ddl");
+                let seq_path = dir_path.join(&seq_file);
+                std::fs::write(&seq_path, seq_ddl)
+                    .with_context(|| format!("failed to write {seq_file}"))?;
+                toc.push_str(&format!("SEQUENCE {seq_key} {seq_file}\n"));
+            }
+
+            let mut ddl = String::new();
+            format::write_create_table(&mut ddl, table);
+            let ddl_file = format!("{}.ddl", table.name);
+            let ddl_path = dir_path.join(&ddl_file);
+            std::fs::write(&ddl_path, &ddl)
+                .with_context(|| format!("failed to write {ddl_file}"))?;
+            toc.push_str(&format!("TABLE {} {ddl_file}\n", table.qualified_name()));
+        }
+    }
+
+    // Data section.
+    if !opts.schema_only {
+        for (idx, table) in tables.iter().enumerate() {
+            let dat_file = format!("{}.dat", idx + 1);
+            let dat_path = dir_path.join(&dat_file);
+
+            let mut data_buf = String::new();
+            format::write_table_data(&mut data_buf, &client, table, opts).await?;
+
+            // Only write data file if there is actual data.
+            if !data_buf.is_empty() {
+                std::fs::write(&dat_path, &data_buf)
+                    .with_context(|| format!("failed to write {dat_file}"))?;
+                toc.push_str(&format!("DATA {} {dat_file}\n", table.qualified_name()));
+            }
+        }
+    }
+
+    // Write toc.dat.
+    let toc_path = dir_path.join("toc.dat");
+    std::fs::write(&toc_path, &toc).context("failed to write toc.dat")?;
+
+    Ok(())
+}
+
+/// Return sequences that any column of this table depends on (via DEFAULT nextval).
+///
+/// Returns a list of (schema, name, create_ddl) tuples.
+async fn get_table_sequences(
+    client: &Client,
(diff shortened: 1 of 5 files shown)


NikolayS:
## REV Code Review — PR #19: directory format (-F directory)

**PR:** feat: directory format dump/restore (-F directory) (#14)
**Author:** NikolayS | +353 / -12 lines
**CI:** ✅ Build & Test passed

---

### BLOCKING ISSUES (1)

**[HIGH] Path traversal in `restore_directory` via untrusted `toc.dat`**
`src/restore/mod.rs` — TOC parsing / dir_path.join

Filenames read from `toc.dat` are joined onto `dir_path` without any sanitization:
```rust
let ddl_path = dir_path.join(ddl_file);   // ddl_file from toc.dat
let dat_path = dir_path.join(dat_file);   // dat_file from toc.dat
```
On Linux/macOS, `Path::join()` with an absolute path **discards** the base — so a toc.dat entry like `TABLE public.users /etc/shadow` would cause the restore to read `/etc/shadow`. A relative traversal (`../../sensitive`) also escapes the dump directory. A crafted dump archive (zip-slip style) could read or — if combined with write paths — overwrite arbitrary files.

**Fix:** Validate that each filename extracted from toc.dat contains no path separators and no leading `/`. Reject or strip any entry that does not pass:
```rust
fn is_safe_filename(s: &str) -> bool {
    !s.contains('/') && !s.contains('\\') && !s.starts_with('.')
}
```
Apply the same check to sequence DDL filenames derived from `seq_name` in `dump_directory`.

---

### NON-BLOCKING (3)

**[MEDIUM] TOCTOU race on output directory creation**
`src/dump/directory_format.rs:24-35`

```rust
if dir_path.exists() {
    // ... check emptiness ...
} else {
    std::fs::create_dir_all(dir_path)  // another process can create it here
```
Between the `exists()` check and `create_dir_all`, another process (or concurrent test run) can create the directory. The resulting error message is confusing ("failed to create directory") rather than "already exists."

**Fix:** Use `create_dir` (not `create_dir_all`) and match on `ErrorKind::AlreadyExists` to get atomic semantics. Then separately check emptiness only on `AlreadyExists`:
```rust
match std::fs::create_dir(dir_path) {
    Ok(_) => {}
    Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => {
        // check empty
    }
    Err(e) => return Err(e).context(...),
}
```

**[MEDIUM] Entire data files loaded into memory**
`src/restore/mod.rs` — `std::fs::read_to_string(&dat_path)`

For tables with millions of rows the `.dat` file could be hundreds of MB or larger. The current code slurps the whole file into a `String`, then clones the data bytes:
```rust
let dat = std::fs::read_to_string(&dat_path)?;
// ...
let data_bytes = bytes::Bytes::from(data.clone());
```
This doubles memory usage (string + bytes clone). A 500 MB table requires ~1 GB RAM.

**Fix:** Stream the file line-by-line or in chunks, or at minimum avoid the `.clone()` by converting the owned string segment to `Bytes` without copying.

**[LOW] Unused variable `_qname` in TOC parser**
`src/restore/mod.rs:68`

```rust
let _qname = parts.next().unwrap_or("").to_string();
```
`_qname` is later used only in the COPY error message but the leading underscore suggests it was intended as intentionally unused — making it confusing. More importantly, the variable is parsed and stored for `data_entries` but for `schema_files` the qname is silently discarded. If we ever need to `--clean` (DROP TABLE before restore), we need the qname for DDL entries too.

**Fix:** Remove the underscore prefix; add a `TODO` comment noting that qname will be needed for `--clean` support for schema entries.

---

### POTENTIAL ISSUES (3)

**[MEDIUM] Empty-table data files are skipped silently**
`src/dump/directory_format.rs:90-96`

```rust
if !data_buf.is_empty() {
    std::fs::write(&dat_path, &data_buf)?;
    toc.push_str(&format!("DATA {} {dat_file}\n", ...));
}
```
Empty tables produce no DATA entry in the TOC, which is correct behavior. However there is no log/notice that the table was skipped. During debugging or verification it's hard to tell whether a table had zero rows or was never reached. Low severity but worth a `// empty table — no data entry written` comment at minimum.

**[MEDIUM] Sequence name used as filename without sanitization (dump side)**
`src/dump/directory_format.rs:74`

```rust
let seq_file = format!("{seq_name}.seq.ddl");
let seq_path = dir_path.join(&seq_file);
std::fs::write(&seq_path, seq_ddl)?;
```
`seq_name` comes from the database catalog (`pg_class.relname`). PostgreSQL allows sequence names with characters that are valid identifiers but problematic as filenames (e.g., names with spaces if quoted). While path traversal is unlikely here (PostgreSQL won't allow `/` in names), names with spaces will create files with spaces, which could cause issues on some systems. Consider replacing non-alphanumeric/non-underscore characters with `_`.

**[LOW] Hardcoded `/tmp/test_dir_dump` path causes test collision**
`tests/pg_dump/t002_pg_dump.rs:1193`

```rust
let out_dir = "/tmp/test_dir_dump";
```
If tests run in parallel (e.g., `cargo test -- --test-threads=N`) two concurrent test runs on the same machine will clobber each other's dump directory. The cleanup uses `let _ = std::fs::remove_dir_all(out_dir)` which silently ignores errors.

**Fix:** Use a unique temporary directory per test run:
```rust
let out_dir = format!("/tmp/test_dir_dump_{}", std::process::id());
```
Or use the `tempfile` crate.

---

### Summary

| Area | Findings | Potential | Filtered |
|------|----------|-----------|---------|
| Security | 1 (path traversal) | 1 (seq name) | — |
| Bugs / Logic | 1 (TOCTOU) | 1 (memory) | — |
| Tests | 1 (temp dir collision) | — | — |
| Guidelines | 1 (unused `_qname`) | — | — |
| **Total** | **3** | **3** | — |

**Verdict: CHANGES REQUESTED** — the path-traversal issue in `restore_directory` is a clear HIGH finding that should be addressed before merge. The TOCTOU and memory issues are non-blocking but worth fixing while the code is fresh.

---

NikolayS:
## REV Re-review (post-fix) — PR #19

**Previous blocking issue (path traversal):** ❌ **Still present — NOT fixed**

After inspecting the current HEAD of `feat/directory-format` (commit `205f4bc`), the path traversal vulnerability is **unchanged**. There is only one commit on this PR, and no filename validation was added to `restore_directory`.

The vulnerable code in `src/restore/mod.rs` (lines 91 and 105) still reads:
```rust
let ddl_path = dir_path.join(ddl_file);  // ddl_file comes from toc.dat — unsanitized
let dat_path = dir_path.join(dat_file);  // dat_file comes from toc.dat — unsanitized
```

On Linux, `Path::join()` with an absolute path silently discards the base. A crafted `toc.dat` with an entry like `TABLE public.users /etc/shadow` causes the restore to read `/etc/shadow`. Relative traversal (`../../etc/passwd`) also works.

**Required fix** (add before line 90 where schema files are used, and before line 104 where data files are used):
```rust
fn is_safe_filename(s: &str) -> bool {
    !s.is_empty()
        && !s.starts_with('/')
        && !s.starts_with('.')
        && !s.contains('/')
        && !s.contains('\\')
}
```
Then validate before every `dir_path.join()` call:
```rust
if !is_safe_filename(ddl_file) {
    anyhow::bail!("unsafe filename in toc.dat: {ddl_file:?}");
}
```

**New issues:** None introduced beyond the original findings.

**Previous non-blocking issues:** Still present (TOCTOU race on dir creation, full-file memory load, `_qname` naming, hardcoded `/tmp/test_dir_dump` in test).

**Verdict: CHANGES REQUESTED ❌** — the blocking path-traversal issue must be fixed before merge.

---
*REV-assisted re-review (AI analysis by [postgres-ai/rev](https://gitlab.com/postgres-ai/rev))*

NikolayS:
## REV Re-review (post-fix) — PR #19

**Previous blocking issue (path traversal):** Fixed correctly ✅

`safe_join` is well-implemented: rejects empty strings, `/`, `\\`, and leading `.`. It is called on **all** archive-derived file paths in the restore path:
- `ddl_path = safe_join(dir_path, ddl_file)` — schema DDL files
- `dat_path = safe_join(dir_path, dat_file)` — data files

`toc.dat` itself uses a hardcoded literal join (`dir_path.join("toc.dat")`), which is correct. No archive-derived path bypasses `safe_join`.

**Dump-side path joins** in `src/dump/directory_format.rs` use raw `.join()` but the filenames are code-generated (numeric indices, hardcoded extensions like `toc.dat`). Technically a table named `"../evil"` could escape the output dir during dump, but the user controls the database at that point — acceptable risk at this stage.

---

**New issues found:**

**1. MEDIUM — Duplicate sequence DDL breaks restore for shared sequences**

In `dump_directory()`, `get_table_sequences()` is called per-table. If two tables share a sequence (e.g., via a shared SERIAL/IDENTITY column, or simply referencing the same sequence), the TOC gets duplicate `SEQUENCE` lines:
```
SEQUENCE public.id_seq id_seq.seq.ddl
SEQUENCE public.id_seq id_seq.seq.ddl   ← duplicate
```
On restore, `schema_files` contains both entries and `CREATE SEQUENCE` is executed twice → `ERROR: relation id_seq already exists`.

Fix: deduplicate by `(seq_schema, seq_name)` in `get_table_sequences()` or track emitted sequences in `dump_directory()` with a `HashSet`.

**2. LOW — Sequence filename collision across schemas**

`seq_file` is built as `format!("{seq_name}.seq.ddl")` (no schema prefix). Two sequences with the same name in different schemas produce the same filename, causing a silent overwrite during dump and a wrong CREATE SEQUENCE DDL being executed for one of them on restore.

Fix: use `format!("{seq_schema}__{seq_name}.seq.ddl")` or similar namespaced pattern.

---

**Verdict:** CHANGES REQUESTED ❌

Path traversal is fixed correctly. But the duplicate-sequence bug (#1) is a correctness issue that will hit any real schema with shared sequences — it breaks restore silently on common patterns. Should be addressed before merge.

---
*REV-assisted re-review*

PR #22, opened 2026-04-04
feat: parallel dump and restore (-j N) (#15)

NikolayS:
Closes #15

Implements parallel pg_dump and pg_restore with -j N workers:
- pg_dump -F directory -j N: dumps N tables concurrently via tokio::spawn + Semaphore
- pg_restore -j N: restores data files in parallel (each worker gets its own DB connection)
- DDL always single-threaded first, data parallel, correctness preserved
- 4 new tests in t004_pg_dump_parallel.rs: parallel_dump, parallel_restore, parallel_dump_inserts, parallel_restore_inserts

Additional improvements bundled in this PR (necessary for tests to pass):
- catalog.rs: query now includes relkind='p' (partitioned parent tables) with partition key/bound/parent metadata
- format.rs: write_create_table handles partition children (PARTITION OF syntax) and parents (PARTITION BY)
- format.rs: SELECT columns cast to ::text to handle custom enum types that tokio-postgres can't decode by OID
- directory_format.rs: emits CREATE SCHEMA IF NOT EXISTS for non-public schemas
- directory_format.rs: emits CREATE TYPE for enum types before table DDL
- restore/mod.rs: handles SCHEMA and TYPE TOC entries; continues on DDL/data errors (matching pg_restore default)
- pg_restore binary: detects directory format by path.is_dir() before reading as file bytes

### DIFF
--- a/src/bin/pg_dump.rs
+++ b/src/bin/pg_dump.rs
@@ -253,6 +253,12 @@ fn main() {
         exclude_tables: Vec::new(),
         no_owner: false,
         no_privileges: false,
+        jobs: cli
+            .jobs
+            .as_deref()
+            .and_then(|s| s.parse::<usize>().ok())
+            .unwrap_or(1)
+            .max(1),
     };
 
     let rt = tokio::runtime::Runtime::new().expect("tokio runtime");
@@ -278,6 +284,17 @@ fn main() {
                     });
                 }
             }
+        }
+        "directory" | "d" => {
+            let output_dir = cli.file.as_deref().unwrap_or_else(|| {
+                eprintln!("pg_dump: directory format requires -f/--file");
+                std::process::exit(1);
+            });
+            rt.block_on(dump::directory_format::dump_directory(&opts, output_dir))
+                .unwrap_or_else(|e| {
+                    eprintln!("pg_dump: {e}");
+                    std::process::exit(1);
+                });
         }
         _ => {
             // Plain format (and unimplemented formats fall back to plain).
--- a/src/bin/pg_restore.rs
+++ b/src/bin/pg_restore.rs
@@ -238,17 +238,35 @@ fn main() {
         }
     };
 
-    let file_bytes = std::fs::read(&filename).unwrap_or_else(|e| {
-        eprintln!("pg_restore: could not open file \"{filename}\": {e}");
-        std::process::exit(1);
-    });
+    let jobs: usize = cli
+        .jobs
+        .as_deref()
+        .and_then(|s| s.parse::<usize>().ok())
+        .unwrap_or(1)
+        .max(1);
 
     let opts = restore::RestoreOptions {
         dbname,
         clean: cli.clean,
+        jobs,
     };
 
     let rt = tokio::runtime::Runtime::new().expect("tokio runtime");
+
+    // Directory format: detect before trying to read as a file.
+    if std::path::Path::new(&filename).is_dir() {
+        rt.block_on(restore::restore_directory(&filename, &opts))
+            .unwrap_or_else(|e| {
+                eprintln!("pg_restore: {e}");
+                std::process::exit(1);
+            });
+        return;
+    }
+
+    let file_bytes = std::fs::read(&filename).unwrap_or_else(|e| {
+        eprintln!("pg_restore: could not open file \"{filename}\": {e}");
+        std::process::exit(1);
+    });
 
     if restore::is_custom_format(&file_bytes) {
         rt.block_on(restore::restore_custom(&file_bytes, &opts))
--- a/src/dump/catalog.rs
+++ b/src/dump/catalog.rs
@@ -26,6 +26,8 @@ pub struct TableInfo {
     pub partition_bound: Option<String>,
     /// Name of the parent partitioned table — Some for partition children.
     pub parent_table: Option<String>,
+    /// Schema of the parent partitioned table — Some for partition children.
+    pub parent_schema: Option<String>,
 }
 
 /// Metadata for a single column.
@@ -79,7 +81,13 @@ pub async fn get_tables(client: &Client, opts: &DumpOptions) -> Result<Vec<Table
                          FROM pg_catalog.pg_inherits i
                          JOIN pg_catalog.pg_class p ON p.oid = i.inhparent
                          WHERE i.inhrelid = c.oid
-                         LIMIT 1) AS parent_table
+                         LIMIT 1) AS parent_table,
+                        (SELECT pn.nspname
+                         FROM pg_catalog.pg_inherits i
+                         JOIN pg_catalog.pg_class p ON p.oid = i.inhparent
+                         JOIN pg_catalog.pg_namespace pn ON pn.oid = p.relnamespace
+                         WHERE i.inhrelid = c.oid
+                         LIMIT 1) AS parent_schema
                  from pg_catalog.pg_class c
                  join pg_catalog.pg_namespace n on n.oid = c.relnamespace
                  where c.relkind in ('r', 'p')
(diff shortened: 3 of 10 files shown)


NikolayS:
## REV Code Review — PR #22: parallel dump/restore (-j N)

**CI:** ✅ Build & Test passed
**Diff:** +589 / -95 lines

---

### BLOCKING ISSUES (2)

#### 🔴 B1 — Cross-schema partition parent generates invalid DDL `[severity: HIGH | confidence: 9]`

**File:** `src/dump/catalog.rs` (subquery `parent_table`) + `src/dump/format.rs` line 25

The catalog query fetches only `p.relname` for the parent table (no `nspname`). `write_create_table` then reconstructs the parent's qualified name by reusing the **child's** schema:

```rust
let parent_qname = format!({}.{}, quote_ident(&table.schema), quote_ident(parent));
```

If child is in `app` schema and parent is in `public` schema, the generated DDL will be:

```sql
CREATE TABLE app.ths_0 PARTITION OF app.ths ...  -- wrong: parent is public.ths
```

Restore will fail with "relation does not exist". Fix: join `pg_namespace` for the parent in the catalog subquery and store `parent_schema` in `TableInfo`.

---

#### 🔴 B2 — File name collisions silently overwrite enum/sequence DDL files `[severity: HIGH | confidence: 8]`

**File:** `src/dump/directory_format.rs` lines 86, 99

Enum type files are named `{type_name}.type.ddl` and sequence files `{seq_name}.seq.ddl` — without schema prefix. If two schemas contain an enum or sequence with the same name (e.g., `public.status` and `billing.status`), the second write silently overwrites the first:

```rust
let type_file = format!("{type_name}.type.ddl");    // no schema in filename
let seq_file  = format!("{seq_name}.seq.ddl");       // same problem
```

The TOC will have two entries pointing at the same file; one type/sequence is lost entirely. Fix: include schema in filename, e.g., `{type_schema}.{type_name}.type.ddl`.

---

### NON-BLOCKING (3)

#### 🟡 N1 — Dead `if/else` branch in `catalog.rs` always calls `get_columns` `[severity: MEDIUM | confidence: 9]`

**File:** `src/dump/catalog.rs` lines 155–160

Both arms of the `if partition_bound.is_some()` branch are identical — `get_columns(client, oid).await?`. The comment says "skip column query" but the code doesn't. This doubles the catalog round-trips for partition children with no benefit. Simplify to a single unconditional call and remove the misleading comment, or implement the skip if it was intentional.

---

#### 🟡 N2 — Parallel restore swallows data errors; process exits 0 even on partial failure `[severity: MEDIUM | confidence: 8]`

**File:** `src/restore/mod.rs` lines 174–175, 204–208

Data errors in both sequential and parallel paths are caught and logged as warnings, but `restore_directory` returns `Ok(())` regardless. The binary exits 0 even when tables are missing or COPY failed. Real `pg_restore` returns exit code 1 when any error occurs (unless `--exit-on-error` is off, but it still sets a warning exit code). 

The parallel path in particular logs the error and then returns `Ok(()))` from the task, so even the panic-propagation on `.await` does not help:

```rust
if let Err(e) = restore_data_file(...) {
    eprintln!("pg_restore: warning: ...");
}
Ok::<(), anyhow::Error>(())   // always Ok
```

Consider tracking a failure flag and returning exit code 1 (or propagating errors).

---

#### 🟡 N3 — Partition ordering sort may be non-deterministic within same nspname/relname tier `[severity: LOW | confidence: 7]`

**File:** `src/dump/catalog.rs` (ORDER BY clause)

The sort key `(partition_bound IS NOT NULL), nspname, relname` correctly puts parents before children, but if a parent table `foo` in `public` schema and its child `foo_0` also in `public` both sort to the same `nspname` bucket, the secondary sort `relname` is alphabetical — this is fine in practice but relies on the parent name alphabetically preceding children (e.g., `ths` < `ths_0`). A child named `alpha_part` with parent `zeta` would be emitted before its parent. The bool first-sort handles the main split, so this is acceptable for the common case, but consider documenting the assumption or adding OID-based ordering as a tiebreaker.

---

### POTENTIAL ISSUES (2)

#### ⚪ P1 — Test: `parallel_restore` only verifies `tplain` row count, not partitioned tables `[severity: MEDIUM | confidence: 8]`

**File:** `tests/pg_dump/t004_pg_dump_parallel.rs`

The parallel restore test inserts 300 rows into `ths` and `tht` but only asserts:
```rust
let count = crate::common::psql_query(&dest_db, "SELECT COUNT(*) FROM tplain");
assert_eq!(count.trim(), "100", ...);
```

No assertions for `ths` (300 rows) or `tht` (300 rows). A regression where partition data is silently dropped would not be caught. Add count checks for partitioned tables.

---

#### ⚪ P2 — No test for `--jobs` with `--schema-only` or `--data-only` flags `[severity: LOW | confidence: 6]`

**File:** `tests/pg_dump/t004_pg_dump_parallel.rs`

The new parallel code branches on `opts.schema_only` and `opts.data_only` but there are no integration tests exercising `pg_dump -Fd -j2 --schema-only` or `pg_restore -j3 --data-only`. These are valid combinations used in production workflows.

---

NikolayS:
## REV Re-review (post-fix) — PR #22

**Commits reviewed:**
- `58b6fe4b` — feat: parallel dump and restore (-j N)
- `41af5a94` — fix: propagate parallel task errors, dedup schema DDL, remove dead code

---

### Previous blocking issues

**B1 — Cross-schema partition DDL: ✅ Fixed correctly**

`TableInfo` now has both `parent_table` and `parent_schema` fields, populated via correlated subqueries in `get_tables`. `write_create_table` uses `parent_schema` (falling back to `table.schema` only if it is `None`, which cannot happen given the catalog query always returns both or neither). `write_copy` / `write_inserts` likewise route partition-child data to the fully-qualified parent. The ordering guarantee (parents before children in the catalog query) is correct: `(partition_bound IS NOT NULL)` sorts `false` < `true` in Postgres, so parents come first. ✅

**B2 — Filename collision (enum/sequence DDL files): ⚠️ Partially fixed**

Sequence files and enum type files now use `{schema}__{name}.{ext}` — the `__` separator survives `safe_join`'s validation (no slash, backslash, or leading dot). ✅

However, **table DDL files still use the bare table name** (`format!("{}.ddl", table.name)` — `directory_format.rs` line 112). If two tables with identical names exist in different schemas (e.g. `myschema.orders` and `legacy.orders`), the second write silently overwrites the first DDL file. Both TOC entries point to the same filename, so on restore only one schema's DDL is applied — the other table is not created. The TOC itself records the schema-qualified name (`TABLE myschema.orders orders.ddl`), so the collision is invisible until restore silently produces a broken schema.

**Fix:** Change line 112 to include the schema prefix:
```rust
let ddl_file = format!("{}__{}.ddl", table.schema, table.name);
```

---

### New issues found

**N1 — `get_table_sequences` misses sequences on partitioned parent tables (LOW, confidence 5)**

`get_table_sequences` filters `WHERE t.relkind = 'r'` (line 285, `directory_format.rs`). Partitioned parents have `relkind = 'p'`, so any sequence dependency on a parent's column is silently skipped. In practice, partitioned tables rarely carry serial columns directly, but it is technically possible (e.g. a `bigserial` on a parent with row-level inserts via triggers). Low severity, but worth a follow-up.

**N2 — Parallel restore swallows data errors (MEDIUM, confidence 7)**

Both sequential and parallel data-restore paths log errors as warnings and continue (`eprintln!("pg_restore: warning: ...")` then fall through). `restore_directory` returns `Ok(())` even on partial data failure, so the process exits 0. This matches real `pg_restore`'s default behavior (non-fatal errors unless `--exit-on-error` is set), and the comments acknowledge this. However:

1. There is no `--exit-on-error` flag implemented yet — users have no way to opt in to fail-fast behavior.
2. In parallel mode (the new code), a connection failure for one worker silently skips all that table's data with no indication beyond a stderr warning.

Not blocking for merge, but should be tracked as a follow-up issue.

---

### Verdict: CHANGES REQUESTED ❌

**B2 is still partially present** — table DDL files don't include the schema prefix, causing silent data loss on restore when two tables share a name across schemas. One-line fix in `directory_format.rs:112`. Everything else (B1, parallelism, error propagation, `safe_join` compatibility) looks good.

---
*REV-assisted re-review*

PR #29, opened 2026-04-04
feat: custom format i32 overflow + end-of-data marker hardening (#20)

### LINKED ISSUE #20: hardening: custom format i32 overflow + end-of-data marker check
Two minor hardening items from REV re-review of PR #18:

1. `compressed.len() as i32` can overflow for tables with >2GB of compressed data. Should use a checked cast or u64 length encoding.

2. `read_next_data_block` does not validate the end-of-data marker byte — a corrupt/truncated archive could cause silent misbehavior instead of a clear error.

Neither blocks correctness for normal-sized tables. Backlog hardening task.

NikolayS:
Fixes #20

## Goal
Two minor hardening items from REV review:
1. `compressed.len() as i32` checked cast — returns io::Error if data exceeds i32::MAX (avoids silent overflow for >2GB tables)
2. `read_next_data_block` now validates end-of-data marker byte and size field — corrupt/truncated archives produce clear errors instead of silent misbehavior

## What changed
- `src/dump/custom_format.rs`: checked cast + marker validation
- `tests/pg_dump/t006_custom_format_hardening.rs`: new hardening tests (not ignored)

## How to verify
`cargo test t006` — all new tests pass

### DIFF
--- a/src/dump/custom_format.rs
+++ b/src/dump/custom_format.rs
@@ -272,11 +272,22 @@ pub fn write_data_block(w: &mut impl Write, dump_id: i32, data: &[u8]) -> io::Re
     // Compress the data.
     let compressed = compress_zlib(data)?;
 
+    // Guard against silent i32 overflow for tables with >2GB of compressed data.
+    let compressed_len = i32::try_from(compressed.len()).map_err(|_| {
+        io::Error::new(
+            io::ErrorKind::InvalidInput,
+            format!(
+                "compressed data length {} exceeds i32::MAX; cannot write data block",
+                compressed.len()
+            ),
+        )
+    })?;
+
     // Block header: type + dump_id.
     w.write_all(&[BLK_DATA])?;
     write_int(w, dump_id)?;
     // Compressed data size.
-    write_int(w, compressed.len() as i32)?;
+    write_int(w, compressed_len)?;
     // Compressed data.
     w.write_all(&compressed)?;
 
@@ -497,12 +508,30 @@ pub fn read_next_data_block(r: &mut impl std::io::Read) -> io::Result<Option<(i3
             let mut decompressed = Vec::new();
             decoder.read_to_end(&mut decompressed)?;
 
-            // Skip the end-of-data marker (BLK_DATA + dump_id + size=0).
+            // Validate the end-of-data marker (BLK_DATA + dump_id + size=0).
+            // A corrupt or truncated archive must produce a clear error.
             let mut marker_type = [0u8; 1];
             r.read_exact(&mut marker_type)?;
-            if marker_type[0] == BLK_DATA {
-                let _id = read_int(r)?;
-                let _sz = read_int(r)?;
+            if marker_type[0] != BLK_DATA {
+                return Err(io::Error::new(
+                    io::ErrorKind::InvalidData,
+                    format!(
+                        "invalid end-of-data marker byte: expected BLK_DATA ({}), got {}",
+                        BLK_DATA, marker_type[0]
+                    ),
+                ));
+            }
+            let _marker_id = read_int(r)?;
+            let marker_size = read_int(r)?;
+            if marker_size != 0 {
+                return Err(io::Error::new(
+                    io::ErrorKind::InvalidData,
+                    format!(
+                        "corrupt archive: end-of-data marker has non-zero size {} (expected 0); \
+                         archive may be truncated",
+                        marker_size
+                    ),
+                ));
             }
 
             Ok(Some((dump_id, decompressed)))
--- a/src/dump/custom_format.rs
+++ b/src/dump/custom_format.rs
@@ -269,10 +269,26 @@ pub fn write_toc_entry(w: &mut impl Write, entry: &TocEntry) -> io::Result<()> {
 ///   - compressed data
 ///   - BLK_DATA + dump_id + size=0 (end-of-data marker for this entry)
 pub fn write_data_block(w: &mut impl Write, dump_id: i32, data: &[u8]) -> io::Result<()> {
-    // Compress the data.
     let compressed = compress_zlib(data)?;
+    write_data_block_compressed(w, dump_id, &compressed)
+}
 
-    // Guard against silent i32 overflow for tables with >2GB of compressed data.
+/// Write a pre-compressed data block for one TOC entry.
+///
+/// This is the inner implementation used by `write_data_block` after
+/// compression. Exposed as `pub(crate)` so that tests can inject an
+/// already-"compressed" buffer of arbitrary length (including one that
+/// exceeds `i32::MAX`) without having to allocate >2 GB of actual memory.
+///
+/// The overflow guard lives here — removing it from this function will
+/// cause `write_data_block_len_overflow_returns_error` to fail.
+pub fn write_data_block_compressed(
+    w: &mut impl Write,
+    dump_id: i32,
+    compressed: &[u8],
+) -> io::Result<()> {
+    // Guard against an incorrect truncating cast for tables whose compressed
+    // output would exceed i32::MAX bytes.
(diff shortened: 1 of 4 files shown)


NikolayS:
## REV review: PR #29 — custom format i32 overflow + end-of-data marker hardening

Reviewed commit: `d66caf4`
Files: `src/dump/custom_format.rs`, `tests/pg_dump/t006_custom_format_hardening.rs`

---

## Fix 1: i32 overflow guard in `write_data_block` ✅

**Correct.** `i32::try_from(compressed.len())` is the right call — clean, idiomatic Rust. Error is properly propagated via `?` and `io::Error::new(InvalidInput, ...)` is appropriate. The error message includes the actual length. No issues here.

**NIT:** Comment says "silent i32 overflow" — in Rust, `as i32` is defined behavior (wrapping/truncation), not silent UB. "incorrect cast" or "truncating cast" would be more precise.

---

## Fix 2: End-of-data marker validation in `read_next_data_block`

**Logic is correct.** The new validation properly errors on:
- Wrong marker type byte (anything != `BLK_DATA`) → `InvalidData`
- Non-zero marker size → `InvalidData`

Both use `io::ErrorKind::InvalidData` which is appropriate for corrupt archive data.

**SUGGESTION (non-blocking):** `_marker_id` is read but not validated (line 524). `write_data_block` always writes the same `dump_id` in the marker, so a mismatch would indicate a spliced/truncated archive. Consider:

```rust
let marker_id = read_int(r)?;
if marker_id != dump_id {
    return Err(io::Error::new(
        io::ErrorKind::InvalidData,
        format!("corrupt archive: end-of-data marker dump_id {} != block dump_id {}", marker_id, dump_id),
    ));
}
```

Not a blocker — size=0 check already catches most truncation. This adds defense-in-depth for spliced-archive corruption.

**SUGGESTION (non-blocking):** The legacy `compressed_size <= 0` early-return path (line 496) is now semantically stale:

```rust
if compressed_size <= 0 {
    // End-of-data marker for a previous entry or zero-size block.
    return Ok(Some((dump_id, Vec::new())));
}
```

In the new model, end-of-data markers are consumed *inside* the positive-size path — they'll never surface here for valid archives (zlib of empty input = 8 bytes, never 0). The comment "End-of-data marker for a previous entry" is now stale/misleading. A corrupt block with `compressed_size <= 0` would silently return `Some` rather than error. Consider replacing with an explicit error for invalid/negative sizes. Not a correctness issue for valid archives, but the comment will confuse future readers.

---

## Test coverage: `t006_custom_format_hardening.rs`

| Test | Valid? | Exercises new code? |
|---|---|---|
| `write_data_block_len_overflow_returns_error` | ⚠️ | **No** — see BLOCKER below |
| `write_data_block_normal_data_roundtrips` | ✅ | Yes |
| `write_data_block_empty_data_roundtrips` | ✅ | Yes — but comment is wrong |
| `read_next_data_block_bad_marker_byte_returns_error` | ✅ | Yes |
| `read_next_data_block_nonzero_marker_size_returns_error` | ✅ | Yes |
| `read_next_data_block_followed_by_eof` | ✅ | Yes |
| `read_next_data_block_valid_marker_ok` | ✅ | Partial duplicate of roundtrip |

**BLOCKER (test validity):** `write_data_block_len_overflow_returns_error` (lines 23–31) does **not** call `write_data_block`. It only tests that `i32::try_from(oversized).is_err()` — a stdlib property, not production code behavior:

```rust
let oversized: usize = (i32::MAX as usize) + 1;
let result = i32::try_from(oversized);
assert!(result.is_err(), ...);
```

**This test passes even if the overflow guard is deleted from `write_data_block`.** It provides false confidence. Options:
1. Add a `#[cfg(test)]` hook / mock `compress_zlib` to inject an oversized buffer, or
2. Replace with a doc comment explaining why this path can't be exercised in unit tests (allocating >2 GB is unrealistic), and mark it `#[ignore]` or remove it

**NIT (test comment):** `write_data_block_empty_data_roundtrips` says "size=0 path" — but `write_data_block(b"")` compresses to 8 bytes (zlib header), so `compressed_size = 8`, not 0. The path label is incorrect. The test is fine behaviorally, just the comment is wrong.

**NIT:** `write_data_block_empty_data_roundtrips` only checks `result.is_ok()` without asserting `id == 7` or `data.is_empty()`, unlike the normal-data roundtrip test. Minor inconsistency.

---

## General / regressions

No unintended behavior changes. Hot path correctness is preserved.

One observation at `src/restore/mod.rs:307`:
```rust
Some((_dump_id, data)) if data.is_empty() => { continue; }
```
This guard was written for the old behavior where `compressed_size <= 0` returned empty data. With the new code, valid archives never produce empty data from `read_next_data_block` (empty input compresses to 8 bytes). The guard is now dead code for valid archives — harmless but worth a follow-up cleanup note.

Code style consistent with codebase. Error messages are informative. `io::ErrorKind::InvalidData` used correctly.

---

## Summary

Core logic of both fixes is **correct and an improvement**. One item to address before merge:

- **BLOCKER:** `write_data_block_len_overflow_returns_error` tests stdlib, not production code — gives false confidence. Rewrite or remove.
- **SUGGESTION:** Validate `_marker_id == dump_id` for defense-in-depth.
- **SUGGESTION:** Tighten/remove the legacy `compressed_size <= 0` path and update its comment.
- **NIT:** Fix the "size=0 path" comment in `write_data_block_empty_data_roundtrips`.

NikolayS (commented):
## Review: custom format i32 overflow + end-of-data marker hardening

CI is green. The logic is sound and the test coverage is good. Two issues to be aware of (one non-blocking, one minor):

---

### 🔴 Non-blocking issue: UB in test's `unsafe` block

`tests/pg_dump/t006_custom_format_hardening.rs`, `write_data_block_len_overflow_returns_error`:

```rust
let oversized: &[u8] = unsafe { std::slice::from_raw_parts(backing.as_ptr(), oversized_len) };
```

The SAFETY comment argues that the guard fires before any memory access, so the backing store is never read past its real size. That's true in practice — `i32::try_from(compressed.len())` only reads the fat-pointer metadata, not the data. **But the Rust language spec says `slice::from_raw_parts` requires the pointer to be valid for `len` reads**, regardless of whether you actually read. Constructing this slice is technically UB.

Simpler safe alternatives that don't need unsafe:
1. Extract the length check into a small helper (`fn check_compressed_len(len: usize) -> io::Result<i32>`) and test that directly.
2. Or expose a `write_data_block_compressed` that takes a `u64` length argument instead of a slice, keeping the real function separate.

In practice this test works fine on all current platforms and Miri would not flag it as a real violation unless the data is accessed, but the unsafe is unnecessary here and sets a bad precedent for tests.

---

### 🟡 Minor: doc comment says `pub(crate)`, actual visibility is `pub`

```rust
/// Exposed as `pub(crate)` so that tests can inject an already-"compressed" buffer …
pub fn write_data_block_compressed(…)
```

The doc says `pub(crate)` but the function is fully `pub`. If the intent was to limit visibility, change `pub fn` → `pub(crate) fn`. If keeping it `pub` for flexibility, update the doc to match.

---

### ✅ Everything else looks correct

- **i32 overflow guard**: `i32::try_from(compressed.len())` + `map_err` → `io::Error` is the right pattern. The guard fires before `write_int` and `write_all`, so no partial write on overflow.
- **End-of-data marker validation**: All three fields (type byte, dump_id, size) are now validated with clear error messages. The `compressed_size <= 0` guard also correctly prevents a `-1 as usize` wrap-around allocation.
- **Test coverage**: Round-trip tests, empty-data case, three corruption scenarios (bad marker byte, wrong dump_id, nonzero size) and EOF sequencing. Comprehensive.
- **Refactoring**: Splitting into `write_data_block` + `write_data_block_compressed` is clean and the doc comment explaining the testability rationale is useful.

No blocking issues. The unsafe test is worth cleaning up but does not need to block merge.

Issue #2, opened 2026-04-03
test: extract PostgreSQL TAP tests and create Rust integration test stubs


NikolayS:
## Sprint 1 — Task 1: Extract Test Cases

Parse the PostgreSQL Perl TAP tests and create Rust integration test stubs (all RED).

### Source files

From the PostgreSQL 18 source (available on system or downloadable):
- `src/bin/pg_dump/t/001_basic.pl`
- `src/bin/pg_dump/t/002_pg_dump.pl`
- `src/bin/pg_dump/t/003_pg_dump_with_server.pl`
- `src/bin/pg_dump/t/004_pg_dump_parallel.pl`
- `src/bin/pg_dump/t/010_dump_connstr.pl`

### Deliverables

Create Rust integration test stubs under `tests/pg_dump/`:
- `t001_basic.rs` — maps to 001_basic.pl test cases
- `t002_pg_dump.rs` — maps to 002_pg_dump.pl test cases
- `t003_pg_dump_with_server.rs` — maps to 003_pg_dump_with_server.pl test cases
- `t004_pg_dump_parallel.rs` — maps to 004_pg_dump_parallel.pl test cases
- `t010_dump_connstr.rs` — maps to 010_dump_connstr.pl test cases

Each test stub should:
1. Have a descriptive name matching the Perl test description
2. Contain a comment with the original Perl test for reference
3. Panic with `todo!("not implemented")` or use `#[ignore]` to stay RED
4. Include the test case number and description

### Acceptance criteria

- Each Perl test case has a corresponding Rust test function
- All stubs compile (even if they all fail/panic)
- Test names are clear and descriptive
- `cargo test -- --ignored` shows all stubs

### Depends on

Issue #1 (scaffold) — needs Cargo project structure first

### Notes

- Fetch Perl source from: https://github.com/postgres/postgres/tree/master/src/bin/pg_dump/t
- Or check if pg source is available at `/usr/share/postgresql/` or download via apt source
- Conventional commit: `test: add integration test stubs from PostgreSQL TAP suite`

PR #30, opened 2026-04-04
feat: --clean/--if-exists/--create flags for dump output

### LINKED ISSUE #23: feat: --clean/--if-exists flags for drop-before-create restore
## Goal
Support `--clean` and `--if-exists` flags in `pg_restore` mode (matching pg_dump/pg_restore behavior).

## Scope
- `--clean`: emit DROP statements before CREATE statements on restore
- `--if-exists`: use `DROP ... IF EXISTS` instead of bare `DROP ...` (only meaningful with `--clean`)
- `--create`: include CREATE DATABASE + \connect in the output

## Tests to unignore
From t003/t010: `run_clean`, `run_clean_if_exists`, `run_createdb`, and related restore tests that exercise these flags.

## References
pg_dump TAP test: t002_pg_dump.rs `run_clean`, `run_clean_if_exists`, `run_createdb`

NikolayS:
Closes #23

## Goal
Implement `--clean`, `--if-exists`, and `--create` flags for pg_dump output mode (matching real pg_dump behavior).

## What changed
- `src/dump/mod.rs`: Added `clean`, `if_exists`, `create_db` fields to `DumpOptions`; `dump_plain` prepends DROP statements when `--clean`, uses `IF EXISTS` when `--if-exists`, emits `CREATE DATABASE` + `\connect` when `--create`
- `src/bin/pg_dump.rs`: Wire CLI flags into `DumpOptions`
- `src/restore/mod.rs`: Added `if_exists` field to `RestoreOptions`; `generate_drop_statements` now respects it
- `src/bin/pg_restore.rs`: Wire `--if-exists` into `RestoreOptions`
- `tests/pg_dump/t002_pg_dump.rs`: Unignored and implemented `run_clean`, `run_clean_if_exists`, `run_createdb`

## How to verify
```
PGHOST=localhost PGPORT=5432 PGUSER=postgres PGPASSWORD=postgres cargo test run_clean
PGHOST=localhost PGPORT=5432 PGUSER=postgres PGPASSWORD=postgres cargo test run_createdb
```

## Test results
Before: 83 passed / 213 ignored
After: **86 passed / 210 ignored** — 3 new tests green, 0 failures

## Screenshots
N/A — CLI tool

### DIFF
--- a/src/bin/pg_dump.rs
+++ b/src/bin/pg_dump.rs
@@ -87,6 +87,14 @@ struct Cli {
     /// Add ON CONFLICT DO NOTHING to INSERT commands
     #[arg(long = "on-conflict-do-nothing")]
     on_conflict_do_nothing: bool,
+
+    /// Dump only tables matching pattern (can be specified multiple times)
+    #[arg(short = 't', long = "table")]
+    table: Vec<String>,
+
+    /// Include CREATE DATABASE + \connect in the output
+    #[arg(short = 'C', long = "create")]
+    create: bool,
 }
 
 /// Build the version string: `pg_dump (pg_plumbing) <version>`.
@@ -250,7 +258,7 @@ fn main() {
 
     let opts = dump::DumpOptions {
         dbname,
-        tables: Vec::new(), // TODO: -t flag not yet plumbed in pg_dump binary
+        tables: cli.table.clone(),
         schema_only: cli.schema_only,
         data_only: cli.data_only,
         inserts: cli.inserts || cli.column_inserts || cli.rows_per_insert.is_some(),
@@ -267,6 +275,9 @@ fn main() {
             .and_then(|s| s.parse::<usize>().ok())
             .unwrap_or(1)
             .max(1),
+        clean: cli.clean,
+        if_exists: cli.if_exists,
+        create_db: cli.create,
     };
 
     let rt = tokio::runtime::Runtime::new().expect("tokio runtime");
--- a/src/bin/pg_dump.rs
+++ b/src/bin/pg_dump.rs
@@ -282,6 +282,19 @@ fn main() {
 
     let rt = tokio::runtime::Runtime::new().expect("tokio runtime");
 
+    // Warn when --clean or --create are used with non-plain formats (they are no-ops).
+    let is_non_plain = matches!(format_str, "custom" | "c" | "directory" | "d");
+    if is_non_plain && cli.clean {
+        eprintln!(
+            "pg_dump: warning: --clean is not yet supported for custom/directory format, ignored"
+        );
+    }
+    if is_non_plain && cli.create {
+        eprintln!(
+            "pg_dump: warning: --create is not yet supported for custom/directory format, ignored"
+        );
+    }
+
     match format_str {
         "custom" | "c" => {
             let bytes = rt.block_on(dump::dump_custom(&opts)).unwrap_or_else(|e| {
--- a/src/bin/pg_restore.rs
+++ b/src/bin/pg_restore.rs
@@ -256,6 +256,7 @@ fn main() {
     let opts = restore::RestoreOptions {
         dbname,
         clean: cli.clean,
+        if_exists: cli.if_exists,
         jobs,
     };
 
--- a/src/dump/mod.rs
+++ b/src/dump/mod.rs
@@ -77,7 +77,7 @@ pub async fn dump_plain(opts: &DumpOptions) -> Result<String> {
     if opts.create_db {
         let dbname = &opts.dbname;
         out.push_str(&format!("CREATE DATABASE \"{dbname}\";\n"));
-        out.push_str(&format!("\\connect {dbname}\n\n"));
+        out.push_str(&format!("\\connect \"{dbname}\"\n\n"));
     }
 
     out.push_str("SET statement_timeout = 0;\n");
--- a/src/dump/mod.rs
+++ b/src/dump/mod.rs
@@ -41,6 +41,12 @@ pub struct DumpOptions {
     pub no_privileges: bool,
     /// Number of parallel dump workers (1 = sequential).
     pub jobs: usize,
+    /// Prepend DROP statements before CREATE statements.
+    pub clean: bool,
+    /// Use DROP ... IF EXISTS (only meaningful with clean).
+    pub if_exists: bool,
+    /// Include CREATE DATABASE + \connect at the start.
+    pub create_db: bool,
 }
 
 /// Dump a database in plain SQL format.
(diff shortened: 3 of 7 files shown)


NikolayS:
## REV Code Review — feat: --clean/--if-exists/--create flags for dump output

Overall direction is correct and the code is clean. Found several real issues worth addressing.

---

### BLOCKERs

**1. `restore_plain` propagates DROP errors; `restore_custom` silently ignores them — inconsistent and dangerous**

In `src/restore/mod.rs` (line ~355):
```rust
// restore_plain — hard failure on DROP errors:
client
    .batch_execute(&drop_stmts)
    .await
    .context("failed to execute clean (DROP) statements")?;
```

But `restore_custom` (line ~287):
```rust
let _ = client.batch_execute(&drops).await; // ignore errors (table may not exist)
```

The real pg_restore ignores DROP errors during `--clean` (table doesn't exist is expected). `restore_plain` should also swallow the error (or at least downgrade to a warning), otherwise `--clean` without `--if-exists` will fail as soon as a table doesn't exist in the target DB — exactly the common case for a fresh restore. Using `--clean --if-exists` works around this, but the flag combination shouldn't be mandatory.

---

**2. `--clean` and `--create` are silently no-ops for custom and directory dump formats**

`dump_custom()` and `dump_directory::dump_directory()` accept `DumpOptions` but neither reads `opts.clean`, `opts.if_exists`, or `opts.create_db`. There's no guard or warning in the CLI either.

In `src/bin/pg_dump.rs`, this passes `clean: cli.clean` into opts, then dispatches to `dump_custom` — but those flags are never used there. User gets no error and no effect. Either implement the behavior or reject with a clear error message:
```
pg_dump: error: --clean is not supported with custom/directory format (use pg_restore --clean instead)
```

---

### SUGGESTION

**3. `--create` output: `\connect` should include quoted dbname**

In `src/dump/mod.rs` (line ~80):
```rust
out.push_str(&format!("CREATE DATABASE \"{dbname}\";\n"));
out.push_str(&format!("\\connect {dbname}\n\n"));
```

`CREATE DATABASE` is quoted, but `\connect` is not. If `dbname` contains spaces or uppercase, `\connect` will fail when replayed with `psql`. Should be:
```rust
out.push_str(&format!("\\connect \"{dbname}\"\n\n"));
```

**4. `main.rs` (unified CLI) missing `--if-exists` validation**

`src/bin/pg_dump.rs` correctly rejects `--if-exists` without `--clean`:
```rust
if cli.if_exists && !cli.clean {
    eprintln!("pg_dump: error: option --if-exists requires option -c/--clean");
```

But `src/main.rs` has no equivalent check in `run_pg_dump`. A user of the unified `pg-plumbing pg-dump` path can pass `--if-exists` alone, and it silently has no effect (since `opts.clean` is false, the code never checks `opts.if_exists`). Same gap in `run_pg_restore`.

---

### NITs

**5. DROP statements in `dump_plain` skip sequences**

`dump_plain` with `--clean` only emits `DROP TABLE`, not `DROP SEQUENCE`. The custom format produces sequences (via `get_sequences_for_table`), and `generate_drop_statements` handles them — but since plain format doesn't emit `CREATE SEQUENCE`, there's nothing to parse. This is a pre-existing limitation, but worth a `// TODO: sequence drops not emitted in plain format` comment.

**6. Tests don't assert ordering: DROP before CREATE**

`run_clean` checks `stdout.contains("DROP TABLE")` and `stdout.contains("CREATE TABLE")` but doesn't verify that DROP appears *before* CREATE. A simple check:
```rust
let drop_pos = stdout.find("DROP TABLE").unwrap();
let create_pos = stdout.find("CREATE TABLE").unwrap();
assert!(drop_pos < create_pos, "DROP must precede CREATE");
```

**7. `run_createdb` doesn't filter by table — dumps entire DB**

```rust
crate::common::run_pg_dump(&["-d", "postgres", "--create"]);
```
No `-t` filter. In a shared CI environment with many test tables, this is slower than needed and the assertion `stdout.contains("CREATE DATABASE")` could in principle match something else. Suggest adding `-t dump_test_simple` for consistency with other tests.

**8. Test `run_clean` doesn't verify DROP matches the specific table**

`stdout.contains("DROP TABLE")` would pass even if a different table is dropped. More robust:
```rust
assert!(stdout.contains("DROP TABLE") && stdout.contains("dump_test_simple"), ...);
```

---

### Summary

The core logic is correct: `--clean`/`--if-exists` flag plumbing is right, `--if-exists`-requires-`--clean` validation exists in `pg_dump.rs`/`pg_restore.rs`, the `generate_drop_statements` refactor is clean. The two BLOCKERs (#1 and #2) are real correctness/UX bugs that should be fixed before merge.

NikolayS (commented):
## Code Review: `--clean` / `--if-exists` / `--create` flags

Overall solid implementation. Flags are correctly wired through all three layers (CLI structs → `DumpOptions`/`RestoreOptions` → logic), and the key constraint (`--if-exists requires --clean`) is validated in both standalone binaries. New tests cover the happy paths and are integrated into CI.

A few issues worth addressing:

---

### 🔴 Bug: `main.rs` `run_pg_restore` missing `--if-exists` validation

`run_pg_dump` (in `main.rs`) correctly bails if `--if-exists` is used without `--clean`:
```rust
if args.if_exists && !args.clean {
    bail!("pg_dump: error: option --if-exists requires option -c/--clean");
}
```

But `run_pg_restore` in the same file has no such check. The standalone `src/bin/pg_restore.rs` binary does validate this, but when calling through the `pg-plumbing pg-restore` subcommand path (via `main.rs`), `--if-exists` without `--clean` is silently accepted and does nothing. Should add:
```rust
if args.if_exists && !args.clean {
    bail!("pg_restore: error: option --if-exists requires option -c/--clean");
}
```

---

### 🟡 Behavior change: `restore_plain` drops errors silently on `--clean`

Before this PR, a `DROP` failure (e.g. permission error, syntax issue) in `restore_plain` was surfaced as a hard error. The PR changes it to `let _ = ...` with a comment saying this matches pg_restore behavior. That reasoning is sound for the "object doesn't exist" case (which is the whole point of plain-restore `--clean`), but it now also swallows real errors like permission denied or connection drops mid-batch. Real `pg_restore --clean` only ignores the "does not exist" class of errors, not all errors.

This is acceptable as an incremental step if the alternative is complex per-statement error handling, but it should be documented as a known divergence, and ideally a follow-up issue filed.

---

### 🟡 `--create` output missing database encoding/tablespace

Real pg_dump `--create` emits:
```sql
CREATE DATABASE "mydb" WITH TEMPLATE = template0 ENCODING = 'UTF8' LOCALE_PROVIDER = libc LOCALE = 'en_US.UTF-8';
```

The PR emits just:
```sql
CREATE DATABASE "mydb";
```

This is a correctness gap — if the dump is restored into a cluster with a different default locale, the behavior diverges silently. At minimum worth a `-- TODO` comment, or querying `pg_database` for the actual encoding.

---

### 🟡 No integration test for `pg_restore --if-exists`

`restore_basic.rs` has `restore_with_clean_flag()` but no test for `--clean --if-exists`. The unit test in `restore/mod.rs` (`generate_drops_without_if_exists`) covers the pure function, but there's no end-to-end test verifying the flag is wired through the restore path. Minor, but symmetrical with what was done for dump.

---

### ✅ What's good

- Flag wiring is correct and complete across all four entry points (`src/bin/pg_dump.rs`, `src/bin/pg_restore.rs`, `main.rs` dump, `main.rs` restore)
- `generate_drop_statements` refactored cleanly — `if_exists` parameter is a straightforward boolean, not a new abstraction
- The `--clean + --data-only` incompatibility check is present in both standalone binaries
- Non-plain format warning for `--clean`/`--create` is a nice touch
- New unit test `generate_drops_without_if_exists` correctly verifies absence of `IF EXISTS`, not just presence of `DROP`
- 3 new integration tests are real (connect to PG, verify output) — not stubs

The `run_pg_restore` missing validation is the only blocker. The rest are improvement opportunities.

PR #31, opened 2026-04-04
feat: table and schema filtering (--table, --exclude-table, --schema, --exclude-schema)

### LINKED ISSUE #24: feat: table and schema filtering (--table, --exclude-table, --schema, --exclude-schema)
## Goal
Support object-level filtering flags:
- `--table <pattern>` / `--exclude-table <pattern>`
- `--schema <pattern>` / `--exclude-schema <pattern>`
- `--exclude-table-data <pattern>`

## Scope
- Pattern matching (fnmatch-style with wildcards)
- Apply filters during dump to limit what goes into the archive
- Apply filters during restore to limit what gets restored

## Tests to unignore
`run_exclude_measurement`, `run_exclude_schema`, `run_exclude_table`, `run_exclude_table_data`, `run_only_measurement`, `run_only_schema`, `run_only_table`, and related tests.

NikolayS:
Closes #24

## Goal
Implement object-level filtering flags for pg_dump (--table, --exclude-table, --schema, --exclude-schema, --exclude-table-data) with fnmatch-style wildcard pattern matching.

## What changed
- `src/dump/filter.rs`: new fnmatch-style pattern matcher (pure Rust, no new deps). Supports `*` (any sequence) and `?` (single char), case-sensitive. Matches against both unqualified name (`foo`) and schema-qualified name (`public.foo`).
- `src/dump/mod.rs`: added `exclude_table_data: Vec<String>` field to `DumpOptions`; filtering logic in `dump_plain` skips data for tables matching `--exclude-table-data` patterns
- `src/dump/catalog.rs`: upgraded schema/table filter logic in `get_tables` to use glob matching (was exact string match)
- `src/bin/pg_dump.rs`: wired `--exclude-table/-T`, `--schema/-n`, `--exclude-schema/-N`, `--exclude-table-data` CLI flags and propagated them into `DumpOptions`
- `src/main.rs`: added `--exclude-table-data` to `PgDumpArgs` and wired it into `DumpOptions`
- `tests/pg_dump/t002_pg_dump.rs`: unignored and implemented `run_exclude_table`, `run_exclude_schema`, `run_only_table`, `run_only_schema`, `run_exclude_table_data`; kept `run_exclude_measurement` and `run_only_measurement` as `#[ignore]` (no measurement/partition table in test schema)

## How to verify
```
PGHOST=localhost PGPORT=5432 PGUSER=postgres PGPASSWORD=postgres cargo test run_exclude
PGHOST=localhost PGPORT=5432 PGUSER=postgres PGPASSWORD=postgres cargo test run_only
```

## Test results
Before: 94 passed / 210 ignored
After: 99 passed / 205 ignored — 5 new tests green, 0 failures

### DIFF
--- a/src/bin/pg_dump.rs
+++ b/src/bin/pg_dump.rs
@@ -92,6 +92,22 @@ struct Cli {
     #[arg(short = 't', long = "table")]
     table: Vec<String>,
 
+    /// Do not dump the named table(s) (can be specified multiple times)
+    #[arg(short = 'T', long = "exclude-table")]
+    exclude_table: Vec<String>,
+
+    /// Dump only the named schema(s) (can be specified multiple times)
+    #[arg(short = 'n', long = "schema")]
+    schema: Vec<String>,
+
+    /// Do not dump the named schema(s) (can be specified multiple times)
+    #[arg(short = 'N', long = "exclude-schema")]
+    exclude_schema: Vec<String>,
+
+    /// Do not dump data for the named table(s) (can be specified multiple times)
+    #[arg(long = "exclude-table-data")]
+    exclude_table_data: Vec<String>,
+
     /// Include CREATE DATABASE + \connect in the output
     #[arg(short = 'C', long = "create")]
     create: bool,
@@ -264,9 +280,10 @@ fn main() {
         inserts: cli.inserts || cli.column_inserts || cli.rows_per_insert.is_some(),
         column_inserts: cli.column_inserts,
         rows_per_insert: cli.rows_per_insert.map(|v| v as u32),
-        schemas: Vec::new(),
-        exclude_schemas: Vec::new(),
-        exclude_tables: Vec::new(),
+        schemas: cli.schema.clone(),
+        exclude_schemas: cli.exclude_schema.clone(),
+        exclude_tables: cli.exclude_table.clone(),
+        exclude_table_data: cli.exclude_table_data.clone(),
         no_owner: false,
         no_privileges: false,
         jobs: cli
--- a/src/dump/catalog.rs
+++ b/src/dump/catalog.rs
@@ -67,7 +67,15 @@ pub async fn get_tables(client: &Client, opts: &DumpOptions) -> Result<Vec<Table
     let table_rows = if opts.tables.is_empty() {
         // Dump all user tables and partitioned tables (no system schemas).
         let mut excluded = vec!["pg_catalog".to_string(), "information_schema".to_string()];
-        excluded.extend(opts.exclude_schemas.clone());
+        // Only push literal (non-glob) patterns into the SQL != ALL() list.
+        // Glob patterns are applied in-memory below via schema_matches_any.
+        let literal_excludes: Vec<String> = opts
+            .exclude_schemas
+            .iter()
+            .filter(|p| !p.contains(['*', '?']))
+            .cloned()
+            .collect();
+        excluded.extend(literal_excludes);
 
         client
             .query(
@@ -147,7 +155,6 @@ pub async fn get_tables(client: &Client, opts: &DumpOptions) -> Result<Vec<Table
         let oid: u32 = row.get::<_, i64>("oid") as u32;
         let schema: &str = row.get("nspname");
         let name: &str = row.get("relname");
-        let relkind: i8 = row.get::<_, i8>("relkind");
         let partition_key: Option<String> = row.get("partition_key");
         let partition_bound: Option<String> = row.get("partition_bound");
         let parent_table: Option<String> = row.get("parent_table");
@@ -165,8 +172,6 @@ pub async fn get_tables(client: &Client, opts: &DumpOptions) -> Result<Vec<Table
         if super::filter::matches_any(&opts.exclude_tables, schema, name) {
             continue;
         }
-
-        let _ = relkind; // used implicitly via partition_key/partition_bound
 
         // Query columns for all tables — used both for DDL (regular/partitioned
         // parents) and for data dumping (partition children).
--- a/src/dump/catalog.rs
+++ b/src/dump/catalog.rs
@@ -154,9 +154,7 @@ pub async fn get_tables(client: &Client, opts: &DumpOptions) -> Result<Vec<Table
         let parent_schema: Option<String> = row.get("parent_schema");
 
         // Schema inclusion filter (supports glob patterns).
-        if !opts.schemas.is_empty()
-            && !super::filter::schema_matches_any(&opts.schemas, schema)
-        {
+        if !opts.schemas.is_empty() && !super::filter::schema_matches_any(&opts.schemas, schema) {
             continue;
         }
         // Schema exclusion filter (supports glob patterns).
--- a/src/dump/catalog.rs
+++ b/src/dump/catalog.rs
@@ -153,13 +153,18 @@ pub async fn get_tables(client: &Client, opts: &DumpOptions) -> Result<Vec<Table
         let parent_table: Option<String> = row.get("parent_table");
         let parent_schema: Option<String> = row.get("parent_schema");
 
-        // Schema inclusion filter.
-        if !opts.schemas.is_empty() && !opts.schemas.iter().any(|s| s == schema) {
+        // Schema inclusion filter (supports glob patterns).
+        if !opts.schemas.is_empty()
+            && !super::filter::schema_matches_any(&opts.schemas, schema)
+        {
             continue;
         }
-        // Table exclusion filter.
-        let fqn = format!("{schema}.{name}");
-        if opts.exclude_tables.iter().any(|t| *t == name || *t == fqn) {
+        // Schema exclusion filter (supports glob patterns).
+        if super::filter::schema_matches_any(&opts.exclude_schemas, schema) {
+            continue;
+        }
+        // Table exclusion filter (supports glob patterns).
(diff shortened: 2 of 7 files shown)


NikolayS (commented):
## PR #31 — Filter flags (-T, -n, -N, --exclude-table-data) + Glob Matching

**Overall:** Solid foundation — the glob engine is correct for the common case, the API surface is wired consistently through both entry points, and Cargo.toml is unchanged (✅ no new deps). But there are two blocking issues: `dump_custom` / `dump_directory` silently ignore `--exclude-table-data`, and the recursive glob matcher has an exponential worst-case that could stack-overflow or hang on adversarial patterns.

---

### 🔴 Blocking: `--exclude-table-data` only works for plain format

**Files:** `src/dump/mod.rs` (dump_custom loop, ~line 209) and `src/dump/directory_format.rs` (~line 146)

`dump_plain` correctly skips data via `filter::matches_any`, but `dump_custom` and `dump_directory` do **not** check `opts.exclude_table_data` at all — they call `write_table_data_to_string` / `write_table_data` unconditionally for every table. This means:

```
pg_dump -F custom --exclude-table-data events -d mydb   # silently dumps events data
pg_dump -F directory --exclude-table-data events -d mydb  # same
```

Fix — mirror the plain-format check in `dump_custom` (inside the `if !opts.schema_only` block):

```rust
for table in &tables {
    let skip_data = filter::matches_any(
        &opts.exclude_table_data,
        &table.schema,
        &table.name,
    );
    if skip_data {
        continue; // include schema TOC entry but skip data block
    }
    // … existing COPY / data block code
}
```

And the same guard in `directory_format.rs` for both the sequential and parallel paths.

---

### 🔴 Blocking: Recursive glob matcher — exponential backtracking / potential stack overflow

**File:** `src/dump/filter.rs`, `glob_match_inner`

The `*`-branch tries every suffix of `t` recursively, which is O(|t|^k) in the number of `*` characters `k`. For patterns like `*a*b*c*d*e` against a long non-matching string the recursion tree explodes. PostgreSQL identifier length is capped at 63 chars so a crash is unlikely in normal use, but an adversarial or misconfigured pattern can cause a hang or stack overflow. Real pg_dump uses the C library `fnmatch(3)` which has the same flaw, but in Rust the default stack is only 8 MB and recursion is cheap to stack-exhaust.

Recommended fix — convert to the standard linear DP approach (no extra deps):

```rust
fn glob_match_inner(p: &[char], t: &[char]) -> bool {
    let (m, n) = (p.len(), t.len());
    // dp[i][j] = pattern[..i] matches text[..j]
    let mut dp = vec![vec![false; n + 1]; m + 1];
    dp[0][0] = true;
    for i in 1..=m {
        if p[i - 1] == '*' { dp[i][0] = dp[i - 1][0]; }
    }
    for i in 1..=m {
        for j in 1..=n {
            dp[i][j] = match p[i - 1] {
                '*' => dp[i - 1][j] || dp[i][j - 1],
                '?' => dp[i - 1][j - 1],
                c   => dp[i - 1][j - 1] && c == t[j - 1],
            };
        }
    }
    dp[m][n]
}
```

This is O(|p| * |t|) time and space — well within budget for 63-char identifiers.

---

### 🔴 Blocking: Glob patterns in `--exclude-schema` are passed raw into SQL `!= ALL()`

**File:** `src/dump/catalog.rs`, `get_tables` all-tables path (~line 70)

When `opts.tables` is empty, `exclude_schemas` patterns are pushed into the SQL `excluded` array:

```rust
excluded.extend(opts.exclude_schemas.clone());
// … WHERE n.nspname != ALL()
```

SQL `!= ALL` does **exact** string comparison, not glob. So `--exclude-schema 'pub*'` sends the literal string `'pub*'` to SQL; `public` is not equal to `'pub*'`, so it passes the SQL filter and goes on to be caught (correctly) by the in-memory glob filter. The result is functionally correct but wastes a round-trip fetching all rows from excluded schemas.

More importantly: this is a silent semantic inconsistency — if someone passes exact schema names the SQL prunes efficiently; if they pass globs the SQL prune does nothing and you fetch potentially millions of rows from schemas that will all be dropped in memory. For large databases this is a real perf problem.

Fix: only add patterns that contain no glob characters to the SQL exclusion list:

```rust
let literal_excludes: Vec<String> = opts.exclude_schemas
    .iter()
    .filter(|p| !p.contains(['*', '?']))
    .cloned()
    .collect();
excluded.extend(literal_excludes);
```

---

### 🟡 Improvement: `--table` (-t) does not support glob patterns; `--exclude-table` does

**File:** `src/dump/catalog.rs`, `get_tables` opts.tables path (~line 107)

When `opts.tables` is non-empty, each entry is resolved via `parse_qualified_name` + exact SQL lookup (`c.relname = `). But `--exclude-table` patterns go through `filter::matches_any` which supports globs. Real pg_dump treats both `-t` and `-T` as POSIX shell-style patterns. The asymmetry should at minimum be documented; ideally `-t 'dump_*'` should also match via glob (switch to a `LIKE`/`~` SQL query or post-filter from a full scan).

---

### 🟡 Improvement: `--table` + `--schema` conflict silently drops tables

**File:** `src/dump/catalog.rs`, in-memory filter loop (~line 153)

When `opts.tables` is non-empty AND `opts.schemas` is non-empty, the SQL fetches the explicitly requested tables but the in-memory schema inclusion filter then silently drops any table whose schema doesn't match. So:

```
pg_plumbing pg-dump -t myschema.mytable --schema public
# → no output, no warning, exit 0
```

Real pg_dump warns when `-t` and `-n` are combined and the table isn't in the named schema. At minimum, add a diagnostic; ideally reject this combination.

---

### 🟡 Improvement: `dump_custom` uses table name only as dep-map key (schema collision)

**File:** `src/dump/mod.rs`, ~line 215

```rust
let schema_id_map: HashMap<String, i32> = schema_entries
    .iter()
    .map(|e| (e.tag.clone(), e.dump_id))  // tag = table.name (unqualified)
    .collect();
```

If two schemas each have a table named `events`, one `dump_id` silently overwrites the other. Use `format!("{}.{}", e.namespace, e.tag)` as the key.

---

### ✅ Good: Glob correctness for common cases

`**` (double-star) correctly collapses to the same semantics as `*` — the recursive pattern handles consecutive `*` by skipping them in the next call's first branch. Empty pattern `glob_match("", "")` → `true`, `glob_match("", "x")` → `false`. All existing tests pass for the documented wildcards.

---

### ✅ Good: All 4 flag combinations wired consistently through both entry points

Both `src/bin/pg_dump.rs` and `src/main.rs` now wire `exclude_table`, `exclude_schema`, `schema`, and `exclude_table_data` into `DumpOptions`. Short flags match pg_dump compat (`-T`, `-n`, `-N`). No flag is missing from either entry point.

---

### 🟡 Test: `run_only_table` negative assertion is potentially flaky

**File:** `tests/pg_dump/t002_pg_dump.rs`, `run_only_table` (~line 1555)

```rust
assert!(
    !stdout.contains("CREATE TABLE public.dump_test_restore"),
    ...
);
```

`dump_test_restore` is created by `setup_restore_test_schema`, which is a separate `OnceLock`. If a test in a concurrent thread calls `setup_restore_test_schema` before (or during) `run_only_table`, that table exists in the DB and the exact-lookup path in catalog.rs correctly excludes it (since we're asking for only `dump_test_simple`). The assertion would pass. But if `setup_restore_test_schema` hasn't run yet, the table doesn't exist and the assertion trivially passes without actually verifying anything. Consider calling `crate::common::setup_restore_test_schema()` explicitly at the start of `run_only_table` to make the negative assertion meaningful.

---

### 🟡 Test: `run_exclude_table` and `run_exclude_schema` missing positive "other tables still appear" assertion

Both tests check that the excluded item is absent, but don't assert that the remaining tables are still present. A bug that excludes *everything* would make both tests pass. Add at least one positive assertion (e.g., `assert!(stdout.contains("PostgreSQL database dump"))` already exists in `run_only_schema`, but the other two tests lack it).

---

### 🟡 Minor: `relkind` fetched as `i8` then immediately silenced

**File:** `src/dump/catalog.rs`, line ~150

```rust
let relkind: i8 = row.get::<_, i8>("relkind");
// …
let _ = relkind; // used implicitly via partition_key/partition_bound
```

PostgreSQL `relkind` is `"char"` (pg type oid 18). `tokio-postgres` maps it to `i8`, which happens to work, but it's fragile. More importantly the comment "used implicitly" is misleading — `relkind` is not actually used at all; parent/child detection relies entirely on `partition_key`/`partition_bound`. Either use the value for a proper `relkind` check or remove the binding entirely.

---

### Summary table

| Area | Status | Severity |
|---|---|---|
| dump_custom/dump_directory ignore --exclude-table-data | 🔴 Bug | Blocking |
| Recursive glob: exponential backtracking | 🔴 Bug | Blocking (adversarial input) |
| Glob patterns leak into SQL != ALL() | 🔴 Bug | Blocking (perf for large DBs) |
| -t vs -T glob asymmetry | 🟡 | Improvement |
| -t + -n silent conflict | 🟡 | Improvement |
| dump_custom dep-map key collision | 🟡 | Improvement |
| run_only_table flaky negative assertion | 🟡 | Test |
| Missing positive assertions in exclude tests | 🟡 | Test |
| relkind i8 / dead binding | 🟡 | Cleanup |
| No new deps (Cargo.toml unchanged) | ✅ | Good |
| All 4 flags wired in both entry points | ✅ | Good |
| Glob correctness for common cases | ✅ | Good |

PR #34, opened 2026-04-04
feat: URI-style connection string support + CLI host/port/username flags

### LINKED ISSUE #27: feat: connection string support in dump/restore (--dbname as connstr, env vars, special chars)
## Goal
Full connection string support matching pg_dump/pg_restore behavior.

## Scope
- `--dbname` accepts a full libpq connection string (e.g. `postgresql://user:pass@host/db?sslmode=require`)
- Special characters in passwords/usernames properly handled (percent-encoding)
- PGPASSWORD, PGHOST, PGPORT, PGUSER env vars respected
- Parallel dump/restore with connection string (`-j` flag)

## Tests to unignore
All tests in `t010_dump_connstr`:
- `full_dump_special_chars`
- `parallel_dump_special_chars`
- `parallel_restore_special_chars`
- `parallel_restore_with_create`
- `pg_dumpall_connstr_ascii_range_1` through `_4`
- `pg_dumpall_connstr_dbname_accepts_connstring`
- `restore_via_psql_cmdline`
- `restore_via_psql_env_vars`

NikolayS:
## Goal
Implements #27 — URI-style connection string passthrough and --host/--port/--username CLI flags.

## What Changed
- `build_conninfo()` now detects `postgresql://` and `postgres://` URIs and passes through unchanged
- New `build_conninfo_with_params(dbname, &ConnParams)` function accepts explicit host/port/user/password overrides
- Added `--host/-h`, `--port/-p`, `--username/-U`, `--password/-W` CLI flags to both PgDumpArgs and PgRestoreArgs
- Disabled clap's auto `-h` for `--help` (conflicted with `--host`); `--help` still works via long form
- Added unit tests for conninfo URI passthrough, key=value passthrough, and explicit params
- Updated t010 stubs with explanatory `#[ignore]` comments explaining why they remain skipped

## How to Verify
```bash
```
Expected: 0 failed. 196 ignored (unchanged).

Closes #27

### DIFF
--- a/src/bin/pg_dump.rs
+++ b/src/bin/pg_dump.rs
@@ -5,6 +5,7 @@
 
 use clap::Parser;
 use pg_plumbing::dump;
+use pg_plumbing::ConnParams;
 
 /// pg_dump dumps a database as a text file or to other formats.
 ///
@@ -13,9 +14,16 @@ use pg_plumbing::dump;
 #[command(
     name = "pg_dump",
     version = pg_dump_version(),
-    about = "pg_dump dumps a database as a text file or to other formats."
+    about = "pg_dump dumps a database as a text file or to other formats.",
+    // Disable clap's automatic -h/--help short flag so that -h can be
+    // used for --host (matching PostgreSQL's pg_dump interface).
+    // --help still works via the long flag.
+    disable_help_flag = true,
 )]
 struct Cli {
+    /// Print help information
+    #[arg(long = "help", action = clap::ArgAction::Help)]
+    help: (),
     /// Database name to dump (positional, alternative to -d)
     #[arg()]
     dbname: Vec<String>,
@@ -119,6 +127,23 @@ struct Cli {
     /// Do not output commands to set privileges (GRANT/REVOKE)
     #[arg(short = 'x', long = "no-acl", alias = "no-privileges")]
     no_acl: bool,
+
+    // ---- Connection options ----
+    /// Database server host or socket directory (overrides PGHOST)
+    #[arg(short = 'h', long = "host")]
+    host: Option<String>,
+
+    /// Database server port (overrides PGPORT)
+    #[arg(short = 'p', long = "port")]
+    port: Option<String>,
+
+    /// Connect as the specified database user (overrides PGUSER)
+    #[arg(short = 'U', long = "username")]
+    username: Option<String>,
+
+    /// Force password prompt (password may also be supplied via PGPASSWORD)
+    #[arg(short = 'W', long = "password")]
+    password: Option<String>,
 }
 
 /// Build the version string: `pg_dump (pg_plumbing) <version>`.
@@ -280,8 +305,19 @@ fn main() {
             .unwrap_or_else(|| "postgres".to_string())
     };
 
+    // Build connection params from CLI flags (they override env vars inside
+    // build_conninfo_with_params).
+    let conn_params = ConnParams {
+        host: cli.host.clone(),
+        port: cli.port.clone(),
+        user: cli.username.clone(),
+        password: cli.password.clone(),
+    };
+    // Resolve the final conninfo now so DumpOptions carries it.
+    let conninfo = pg_plumbing::build_conninfo_with_params(&dbname, &conn_params);
+
     let opts = dump::DumpOptions {
-        dbname,
+        dbname: conninfo,
         tables: cli.table.clone(),
         schema_only: cli.schema_only,
         data_only: cli.data_only,
--- a/src/bin/pg_restore.rs
+++ b/src/bin/pg_restore.rs
@@ -5,6 +5,7 @@
 
 use clap::Parser;
 use pg_plumbing::restore;
+use pg_plumbing::ConnParams;
 
 /// pg_restore restores a PostgreSQL database from an archive
 /// created by pg_dump.
@@ -14,9 +15,17 @@ use pg_plumbing::restore;
 #[command(
     name = "pg_restore",
     version = pg_restore_version(),
-    about = "pg_restore restores a PostgreSQL database from an archive created by pg_dump."
+    about = "pg_restore restores a PostgreSQL database from an archive created by pg_dump.",
+    // Disable clap's automatic -h/--help short flag so that -h can be
+    // used for --host (matching PostgreSQL's pg_restore interface).
+    // --help still works via the long flag.
+    disable_help_flag = true,
 )]
 struct Cli {
+    /// Print help information
+    #[arg(long = "help", action = clap::ArgAction::Help)]
+    help: (),
+
     /// Target database name or connection string.
     #[arg(short = 'd', long = "dbname")]
     dbname: Option<String>,
@@ -76,6 +85,23 @@ struct Cli {
     /// Archive file(s) to restore (positional).
     #[arg()]
     filenames: Vec<String>,
+
+    // ---- Connection options ----
+    /// Database server host or socket directory (overrides PGHOST)
+    #[arg(short = 'h', long = "host")]
+    host: Option<String>,
+
+    /// Database server port (overrides PGPORT)
+    #[arg(short = 'p', long = "port")]
+    port: Option<String>,
+
+    /// Connect as the specified database user (overrides PGUSER)
+    #[arg(short = 'U', long = "username")]
+    username: Option<String>,
+
+    /// Force password prompt (password may also be supplied via PGPASSWORD)
+    #[arg(short = 'W', long = "password")]
+    password: Option<String>,
 }
 
 /// Build the version string: `pg_restore (pg_plumbing) <version>`.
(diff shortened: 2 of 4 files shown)


NikolayS (commented):
## REV Review: PR #34 — URI connection string support + CLI connection flags

**LGTM** ✅ — merge with minor follow-up items noted below

---

### Strengths

1. **URI passthrough is correct.** `postgresql://` and `postgres://` detection is clean and handles both prefix variants. tokio-postgres supports URIs natively, so pass-through is the right approach — no re-parsing needed.

2. **`build_conninfo_with_params` design is solid.** The three-tier priority chain (explicit params → env vars → compiled-in defaults) matches libpq semantics exactly. The `.clone().or_else(|| env::var(...).ok()).unwrap_or_else(|| ...)` pattern is idiomatic Rust.

3. **Backward compatibility preserved.** `build_conninfo()` is kept as a zero-params convenience wrapper delegating to `build_conninfo_with_params`. Existing callers in `dump/mod.rs` and `restore/mod.rs` are unaffected.

4. **`-h` / `--help` conflict resolved correctly.** `disable_help_flag = true` + manual `#[arg(long = "help", action = clap::ArgAction::Help)]` is the documented clap approach. Both binaries handled consistently. The in-code comment explaining the rationale is helpful.

5. **t010 `#[ignore]` comments are now informative.** The original bare `#[ignore]` told readers nothing. The new comments correctly distinguish between "needs LATIN1 DB" (infra gap) and "needs pg_dumpall implementation" (code gap). The cross-reference to unit tests in `pg_dumpall_connstr_dbname_accepts_connstring` is a nice touch.

6. **Unit tests exercise the real logic.** The four tests in `src/lib.rs` cover URI passthrough, key=value passthrough, full explicit params, and password inclusion. They're isolated from env vars by supplying all params explicitly — good CI hygiene.

---

### Issues

**Non-blocking (nits):**

**1. Double-processing is safe but architecturally fragile**
`main()` calls `build_conninfo_with_params()` and stores the result (a `key=value` string) in `DumpOptions.dbname`. Then `dump_plain()` / `dump_custom()` / `restore_*()` call `build_conninfo(&opts.dbname)` on that already-processed string. This works today because the `=` passthrough catches it — but it's a hidden invariant. A future caller who passes `ConnParams` directly into `DumpOptions` without pre-processing would get the correct conninfo silently discarded, or vice versa.

Suggested fix (non-blocking, can be done in a follow-up): either rename `DumpOptions.dbname` to `DumpOptions.conninfo` to signal it's already resolved, or thread `ConnParams` through to the module functions and build conninfo once there.

**2. `conninfo_params_override_produce_password` test is a near-duplicate**
`conninfo_bare_dbname_uses_params` and `conninfo_params_override_produce_password` test essentially the same code path with slightly different values. Neither covers the env-var fallback path (what happens when `host` is `None` but `PGHOST` is set). A test using `std::env::set_var` under a mutex — or a helper that temporarily overrides one field — would strengthen confidence in the fallback chain.

**3. `--password/-W` semantics diverge from real pg_dump**
Real pg_dump's `-W` is a boolean flag that *forces a prompt* — it doesn't accept a value. Here it accepts `Option<String>`, silently accepting a password on the command line. This is arguably better for automation (no TTY required), but it's an intentional divergence that should be documented in the `--help` string (currently says "Force password prompt…" which is misleading). Low-priority since it's more capable than the original, but worth noting.

**4. `port` typed as `Option<String>` instead of `Option<u16>`**
Minor: port is a number. Accepting it as a string means `--port=banana` is silently forwarded and will fail at connection time with a confusing libpq error rather than a clean CLI validation error. Easy to change to `Option<u16>` — though it'd need a `.to_string()` in `ConnParams` construction. Non-blocking.

---

### Verdict

**Merge it.** The core logic is correct, the interface is clean, CI is green, and the architectural nit (#1) is a known pattern in this codebase (not introduced by this PR). File issues for the four items above if you want them tracked, but none block ship.

NikolayS (commented):
**Summary:** URI/key=value passthrough and -h/-p/-U/-W CLI flags are well-structured, but the PR introduces a blocking regression: `DumpOptions.dbname` now holds a fully-expanded conninfo string, which breaks the `--create` path (emits `CREATE DATABASE "host=localhost port=5432 ..."`) and leaks plaintext passwords into dump output.

**Verdict:** REQUEST_CHANGES

---

## ✅ What's good

- **URI passthrough logic is correct.** Checking `postgresql://` / `postgres://` prefix before the `=` heuristic is the right order of precedence; tokio-postgres handles both forms natively.
- **`ConnParams::default()` backward-compat shim** keeps `build_conninfo()` callers untouched — clean API evolution.
- **`disable_help_flag` + manual `--help`** is the correct clap pattern for reclaiming `-h`; works as expected (`-h localhost --help` shows help, `-h` alone is treated as `--host`).
- **Double-processing is safe** for the URI/key=value cases because the `contains('=')` guard short-circuits the second `build_conninfo` call in `restore/mod.rs` and `dump/mod.rs`. Not pretty, but correct.
- **`#[ignore]` comments** are now meaningfully differentiated (LATIN1 vs. pg_dumpall not-yet-implemented) — much better than the previous bare `#[ignore]`.
- All 115 integration tests + 7 restore tests pass.

---

## ⚠️ Issues

### 🔴 BLOCKING — `--create` emits broken SQL and leaks password

The `dump/mod.rs` code that handles `--create` uses `opts.dbname` directly as the database name:

```rust
let dbname = &opts.dbname;   // now a full conninfo string
// ...
out.push_str(&format!("CREATE DATABASE \"{dbname}\";\n"));
out.push_str(&format!("\\connect \"{dbname}\"\n\n"));
```

After this PR, `opts.dbname` is the fully-expanded conninfo, not a bare DB name. Running:

```
pg_dump -d postgres --create
```

now outputs:

```sql
CREATE DATABASE "host=localhost port=5432 user=postgres dbname=postgres password=postgres";
\connect "host=localhost port=5432 user=postgres dbname=postgres password=postgres"
```

Two problems in one:
1. **Wrong database name** — the generated SQL is semantically wrong; restoring this dump would create a database with a connstring as its name.
2. **Password leaked in plaintext** — the PGPASSWORD / `-W` value appears verbatim in the dump file, and also in the directory-format TOC (`;  dbname: host=... password=...`).

The existing `run_createdb` test misses this because it only asserts `stdout.contains("CREATE DATABASE")` without checking the quoted name.

**Fix:** `DumpOptions` (and `RestoreOptions`) should carry **both** the bare `dbname: String` (for DDL generation) and the resolved `conninfo: String` (for connection). Internal callers in `dump/mod.rs` / `restore/mod.rs` must use `opts.conninfo` to connect and `opts.dbname` for DDL — or introduce a `DumpTarget` enum.

---

### 🟡 NON-BLOCKING — No quoting of special characters in conninfo values

`build_conninfo_with_params` interpolates host/user/password/dbname directly into the key=value string:

```rust
format!("host={host} port={port} user={user} dbname={dbname}")
```

libpq key=value format requires values containing spaces, single quotes, or backslashes to be quoted with single quotes and escaped (`\'` and `\\`). Without quoting:

- `-h "my host"` → `host=my host` → libpq parses `host=my` then treats `host` as an unknown key
- `-U "my=user"` → `user=my=user` → corrupted parse
- `-W "p'w"` → `password=p'w` → parse error

This was a latent bug in the original `build_conninfo` (env vars rarely contain spaces), but now that users pass values directly from the CLI, it's more easily triggered. The fix is a small `quote_conninfo_value(v: &str) -> String` helper.

---

### 🟡 NON-BLOCKING — `-h` silently ignored when a URI is passed as dbname

```
pg_dump -h different_host "postgresql://myhost/mydb"
```

The URI short-circuits before CLI params are applied, so `-h different_host` is silently dropped. Real pg_dump raises an error when you mix URI and individual connection flags. At minimum, emit a warning.

---

### 🟡 NON-BLOCKING — Port is `Option<String>`, not validated as numeric

`--port` accepts any string; `-p notanumber` produces `port=notanumber` in the conninfo and fails with an opaque libpq error. Using `value_parser = clap::value_parser!(u16)` or a simple upfront `str::parse::<u16>()` check would give a much clearer error, matching upstream pg_dump behavior.

---

### 🟡 NON-BLOCKING — Test redundancy and missing cases

- `conninfo_bare_dbname_uses_params` and `conninfo_params_override_produce_password` test essentially the same code path; consolidating them frees room for more useful cases.
- Missing tests: (a) env-var fallback when all `ConnParams` fields are `None`, (b) `None`-password → no `password=` suffix in output, (c) URI passthrough takes priority over non-empty `ConnParams` (i.e., `-h` is ignored for URIs — document or test this intent).

---

## 💡 Suggestions

1. **Split `DumpOptions.dbname`** into `dbname: String` + `conninfo: String` now; the struct is small and callers are few. This cleanly resolves the `--create` regression without structural churn.

2. **Add a `quote_conninfo_value` helper:**
```rust
fn quote_conninfo_value(v: &str) -> String {
    if v.chars().any(|c| c == ' ' || c == '\\' || c == '\'' || c == '\t') {
        format!("'{}'", v.replace('\\', "\\\\").replace('\'', "\\'"))
    } else {
        v.to_string()
    }
}
```
Then use it for all four fields in `build_conninfo_with_params`.

3. **Strengthen the `run_createdb` test** to assert the quoted name is exactly `"postgres"`, not a conninfo string — this would have caught the regression automatically.

4. **Consider redacting passwords from error messages.** The `failed to connect to database "{}"` format string currently prints the full conninfo including `password=...` to stderr.

PR #60, opened 2026-04-04
feat: emit TS objects, AMs, aggregates, casts, collations, conversions, languages (closes #53)

### LINKED ISSUE #53: t002: emit text search objects, access methods, aggregates, casts, collations, conversions, languages
## Goal
Enable t002 tests for text search objects, access methods, aggregates, casts, collations, conversions, and procedural languages.

### Text Search
- `create_ts_configuration` — CREATE TEXT SEARCH CONFIGURATION dump_test.alt_ts_conf1
- `create_ts_dictionary` — CREATE TEXT SEARCH DICTIONARY dump_test.alt_ts_dict1
- `create_ts_parser` — CREATE TEXT SEARCH PARSER dump_test.alt_ts_prs1
- `create_ts_template` — CREATE TEXT SEARCH TEMPLATE dump_test.alt_ts_temp1
- `alter_ts_config_owner` — ALTER TEXT SEARCH CONFIGURATION ... OWNER TO
- `alter_ts_dict_owner` — ALTER TEXT SEARCH DICTIONARY ... OWNER TO
- `alter_ts_configuration_mapping` — ALTER TEXT SEARCH CONFIGURATION ... ADD MAPPING
- `comment_on_text_search_objects`

### Access Methods
- `create_access_method` — CREATE ACCESS METHOD gist2
- `create_access_method_table_am` — CREATE ACCESS METHOD regress_test_table_am
- `create_table_am` — CREATE TABLE using custom AM
- `create_matview_am` — CREATE MATERIALIZED VIEW using custom AM

### Aggregates, Casts, Collations, Conversions
- `create_aggregate` — CREATE AGGREGATE dump_test.newavg
- `create_cast` — CREATE CAST FOR timestamptz
- `create_collation` — CREATE COLLATION test0 FROM "C"
- `create_collation_icu` — CREATE COLLATION icu_collation (ICU)
- `alter_collation_owner` — ALTER COLLATION test0 OWNER TO
- `comment_on_collation` — COMMENT ON COLLATION test0
- `create_conversion` — CREATE CONVERSION dump_test.test_conversion
- `comment_on_conversion`

### Languages
- `create_language` — CREATE LANGUAGE pltestlang
- `create_function_pltestlang_handler` — CREATE FUNCTION pltestlang_call_handler
- `alter_language_owner` — ALTER PROCEDURAL LANGUAGE pltestlang OWNER TO
- `drop_extension_function_language` — DROP EXTENSION/FUNCTION/LANGUAGE in --clean

## Acceptance

NikolayS:
## Goal
Enable t002 tests for TS objects, AMs, aggregates, casts, collations, conversions, languages. Closes #53.

## What changed
- `src/dump/catalog.rs`: catalog queries for TS configs/dicts/parsers/templates, AMs, aggregates, casts, collations, conversions, languages
- `src/dump/format.rs`: emit functions for each new object type
- `src/dump/mod.rs`: wire new emit functions into dump pipeline
- `tests/common/mod.rs`: setup SQL for new objects
- `tests/pg_dump/t002_pg_dump.rs`: all 23 target tests enabled with real assertions

## How to verify
```bash
cd /tmp/pg_plumbing-issue53
export PATH="$HOME/.cargo/bin:$PATH"
```

## Test results (before PR)
- 213 passed, 0 failed, 98 ignored (was 190/0/121)
- All 23 target tests: PASS

## CI
CI must be green before merge.

### DIFF
--- a/src/dump/catalog.rs
+++ b/src/dump/catalog.rs
@@ -37,6 +37,9 @@ pub struct TableInfo {
     pub rls_enabled: bool,
     /// Index name marked for CLUSTER ON, if any.
     pub cluster_index: Option<String>,
+    /// Access method name when non-heap (e.g. `regress_test_table_am`).
+    /// None for regular heap tables.
+    pub am_name: Option<String>,
 }
 
 /// Metadata for a single column.
@@ -184,10 +187,14 @@ pub async fn get_tables(client: &Client, opts: &DumpOptions) -> Result<Vec<Table
                          FROM pg_catalog.pg_index idx
                          JOIN pg_catalog.pg_class ci ON ci.oid = idx.indexrelid
                          WHERE idx.indrelid = c.oid AND idx.indisclustered = true
-                         LIMIT 1) AS cluster_index
+                         LIMIT 1) AS cluster_index,
+                        CASE WHEN am.amname IS NOT NULL AND am.amname <> 'heap'
+                             THEN am.amname
+                        END AS am_name
                  from pg_catalog.pg_class c
                  join pg_catalog.pg_namespace n on n.oid = c.relnamespace
                  join pg_catalog.pg_roles r on r.oid = c.relowner
+                 left join pg_catalog.pg_am am on am.oid = c.relam
                  where c.relkind in ('r', 'p')
                    and n.nspname != all($1)
                  order by
@@ -229,11 +236,15 @@ pub async fn get_tables(client: &Client, opts: &DumpOptions) -> Result<Vec<Table
                              FROM pg_catalog.pg_index idx
                              JOIN pg_catalog.pg_class ci ON ci.oid = idx.indexrelid
                              WHERE idx.indrelid = c.oid AND idx.indisclustered = true
-                             LIMIT 1) AS cluster_index
+                             LIMIT 1) AS cluster_index,
+                            CASE WHEN am.amname IS NOT NULL AND am.amname <> 'heap'
+                                 THEN am.amname
+                            END AS am_name
                      from pg_catalog.pg_class c
                      join pg_catalog.pg_namespace n
                        on n.oid = c.relnamespace
                      join pg_catalog.pg_roles r on r.oid = c.relowner
+                     left join pg_catalog.pg_am am on am.oid = c.relam
                      where c.relkind in ('r', 'p')
                        and n.nspname = $1
                        and c.relname = $2
@@ -260,6 +271,7 @@ pub async fn get_tables(client: &Client, opts: &DumpOptions) -> Result<Vec<Table
         let parent_schema: Option<String> = row.get("parent_schema");
         let rls_enabled: bool = row.get("relrowsecurity");
         let cluster_index: Option<String> = row.get("cluster_index");
+        let am_name: Option<String> = row.get("am_name");
 
         // Schema inclusion filter (supports glob patterns).
         if !opts.schemas.is_empty() && !super::filter::schema_matches_any(&opts.schemas, schema) {
@@ -294,6 +306,7 @@ pub async fn get_tables(client: &Client, opts: &DumpOptions) -> Result<Vec<Table
             parent_schema,
             rls_enabled,
             cluster_index,
+            am_name,
         });
     }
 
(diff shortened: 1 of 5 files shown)


NikolayS:
## REV Review — PR #60 (feat: TS objects, AMs, aggregates, casts, collations, conversions, languages)

**Overall:** Good coverage, catalog queries are mostly correct, OnceLock pattern used properly. Two blocking issues need fixing before merge.

---

### 🔴 BLOCKING

#### 1. `CREATE PROCEDURAL LANGUAGE` emitted before handler function (unrestorable output)

In `src/dump/mod.rs`, the language loop is inserted **before** the function loop:

```
// Emit procedural languages (before functions that use them).
for lang in &languages { ... CREATE PROCEDURAL LANGUAGE pltestlang HANDLER pltestlang_call_handler ... }

// Emit functions and procedures.
for func in &functions { ... CREATE FUNCTION pltestlang_call_handler() ... }
```

The comment says "before functions" but the dependency is the other way: `CREATE PROCEDURAL LANGUAGE` requires the handler function to already exist. A dump produced by this code cannot be restored — the `CREATE LANGUAGE` statement will fail with `function pltestlang_call_handler() does not exist`.

**Fix:** Move the language loop to **after** the function loop (and flip the comment). Real pg_dump outputs functions first, then procedural languages.

**Why tests miss this:** The tests only check that the strings appear somewhere in the output (`stdout.contains(...)`), not that `CREATE FUNCTION` precedes `CREATE LANGUAGE`.

---

#### 2. `create_table_am` test does not assert the `USING` clause

The test for custom table access methods:

```rust
fn create_table_am() {
    // ...
    assert!(
        stdout.contains("regress_pg_dump_table_am"),
        "output should contain table regress_pg_dump_table_am:\n{stdout}"
    );
}
```

This assertion only checks the **table name** appears in output — any reference anywhere would pass it (e.g. a comment). It does **not** verify the `USING regress_test_table_am` clause that is the entire point of the table AM feature.

**Fix:** Add a second assertion:
```rust
assert!(
    stdout.contains("USING regress_test_table_am"),
    "regress_pg_dump_table_am should include USING clause:\n{stdout}"
);
```

---

### 🟡 NON-BLOCKING

#### 3. `write_create_collation` / `write_alter_collation_owner` / `write_comment_on_collation` drop the schema

All three collation format functions use `quote_ident(&col.name)` without a schema prefix. If a collation lives in a non-`public` schema, the generated DDL would be unqualified and potentially target the wrong collation or fail. Example: a collation in `dump_test` would dump as `CREATE COLLATION my_coll (`... but should be `dump_test.my_coll`.

The `CollationInfo` struct has a `qualified_name()` method — just use it.

```rust
// Current (wrong for non-public schemas):
out.push_str(&format!("CREATE COLLATION {} (", quote_ident(&col.name)));

// Fixed:
out.push_str(&format!("CREATE COLLATION {} (", col.qualified_name()));
```

Tests don't catch this because `test0` / `icu_collation` are created without a schema (go to `public`).

---

#### 4. `write_create_access_method` doesn't qualify the handler function schema

```rust
quote_ident(&am.handler_func)   // handler_schema is fetched but silently ignored
```

For the test cases (`gisthandler`, `heap_tableam_handler`) this is fine since they live in `pg_catalog` and PG resolves them unqualified. But a user-written AM handler in a custom schema would produce broken DDL. The `handler_schema` field is fetched from the catalog but never used in the format function. Apply the same schema-qualification pattern used for all other functions in this PR.

---

#### 5. N+1 query in `get_ts_configs`

For each TS config, a second round-trip fires to fetch `prs_oid`:

```rust
let prs_oid: i64 = {
    let pr = client.query_one("SELECT p.oid::bigint AS prs_oid FROM pg_catalog.pg_ts_config c JOIN pg_catalog.pg_ts_parser p ON p.oid = c.cfgparser WHERE c.oid = ::bigint::oid", ...)...
};
```

The main query already joins `pg_ts_parser p ON p.oid = c.cfgparser` — just add `p.oid::bigint AS prs_oid` to the SELECT list and drop the inner query entirely.

---

### ℹ️ INFO

- **OnceLock pattern:** Correct. `setup_issue53_schema` uses `static INIT: OnceLock<()>` and avoids test poisoning. ✅
- **Catalog queries:** `pg_ts_config`, `pg_ts_dict`, `pg_ts_parser`, `pg_ts_template`, `pg_am`, `pg_aggregate`, `pg_cast`, `pg_collation`, `pg_conversion`, `pg_language` — all correct joins. ✅
- **`oid > 16383` filter** for AMs, casts, languages: appropriate user-object filter. ✅
- **Glob exclude patterns** correctly excluded from the `!= all()` SQL param (literal-only). ✅
- **Copyright:** 2026. ✅
- **`write_create_ts_template` trailing comma:** Correct — INIT line has trailing comma, LEXIZE does not. ✅
- **`USING` clause after `PARTITION BY`** in `write_create_table`: Correct per PostgreSQL grammar (USING follows PARTITION BY). ✅

---

**VERDICT: CHANGES NEEDED** — Fix the language/function emit order (#1) and the `create_table_am` test assertion (#2) before merge.

NikolayS:
Fixed both blocking issues: (1) language loop moved after function loop — CREATE PROCEDURAL LANGUAGE now emits after handler CREATE FUNCTION; (2) create_table_am test now asserts USING clause present. All tests passing.

PR #62, opened 2026-04-04
feat: partitioned tables, types, domains, operators (closes #54)

### LINKED ISSUE #54: t002: partitioned tables (measurement), generated/identity columns, domains, types, operators

NikolayS:
Implements t002 test stubs for partitioned tables (measurement), generated/identity columns, domains, composite/range/enum types, operators, operator families, and operator classes.

Closes #54

### DIFF
--- a/src/dump/catalog.rs
+++ b/src/dump/catalog.rs
@@ -59,6 +59,10 @@ pub struct ColumnInfo {
     pub storage_override: Option<char>,
     /// Column-level `n_distinct` option from `attoptions`, if set.
     pub n_distinct: Option<f64>,
+    /// Generated column expression: Some(expr) when attgenerated = 's' (STORED).
+    pub generated_expr: Option<String>,
+    /// Identity column: Some('a') = ALWAYS, Some('d') = BY DEFAULT, None = not identity.
+    pub identity: Option<char>,
 }
 
 /// Metadata for a constraint.
@@ -391,6 +395,13 @@ pub async fn get_sequences(client: &Client, opts: &DumpOptions) -> Result<Vec<Se
                ON a.attrelid = d.refobjid AND a.attnum = d.refobjsubid
              WHERE s.relkind = 'S'
                AND n.nspname != all($1)
+               -- Exclude identity sequences (these are emitted as ALTER TABLE ... ADD GENERATED ... AS IDENTITY)
+               AND NOT EXISTS (
+                   SELECT 1 FROM pg_catalog.pg_depend di
+                   WHERE di.objid = s.oid
+                     AND di.classid = 'pg_catalog.pg_class'::regclass
+                     AND di.deptype = 'i'
+               )
              ORDER BY n.nspname, s.relname",
             &[&excluded],
         )
@@ -649,7 +660,9 @@ async fn get_columns(client: &Client, table_oid: u32) -> Result<Vec<ColumnInfo>>
                     a.attstattarget::int4 as attstattarget, \
                     a.attstorage::text as attstorage, \
                     t.typstorage::text as typstorage, \
-                    COALESCE(array_to_string(a.attoptions, ','), '') as attoptions \
+                    COALESCE(array_to_string(a.attoptions, ','), '') as attoptions, \
+                    a.attgenerated::text as attgenerated, \
+                    a.attidentity::text as attidentity \
              from pg_catalog.pg_attribute a \
              left join pg_catalog.pg_attrdef d \
                on d.adrelid = a.attrelid and d.adnum = a.attnum \
@@ -669,6 +682,8 @@ async fn get_columns(client: &Client, table_oid: u32) -> Result<Vec<ColumnInfo>>
         let attstorage: &str = row.get("attstorage");
         let typstorage: &str = row.get("typstorage");
         let attoptions: &str = row.get("attoptions");
+        let attgenerated: &str = row.get("attgenerated");
+        let attidentity: &str = row.get("attidentity");
 
         // Determine storage override: only emit if different from type default.
         let storage_override = if attstorage != typstorage {
@@ -688,14 +703,40 @@ async fn get_columns(client: &Client, table_oid: u32) -> Result<Vec<ColumnInfo>>
             _ => None,
         };
 
+        // Generated column: attgenerated = 's' means STORED.
+        // The default_expr holds the generation expression.
+        let generated_expr = if attgenerated == "s" {
+            row.get("default_expr")
+        } else {
+            None
+        };
+
+        // For generated columns, clear default_expr so we don't also emit DEFAULT.
+        let default_expr: Option<String> = if attgenerated == "s" {
+            None
+        } else {
+            row.get("default_expr")
+        };
+
+        // Identity column: 'a' = ALWAYS, 'd' = BY DEFAULT.
+        let identity = if attidentity == "a" {
+            Some('a')
+        } else if attidentity == "d" {
+            Some('d')
+        } else {
+            None
+        };
+
         columns.push(ColumnInfo {
             name: row.get("attname"),
             type_name: row.get("type_name"),
             not_null: row.get("attnotnull"),
-            default_expr: row.get("default_expr"),
+            default_expr,
             statistics, // Some(n≥0) if explicitly set by user, None = system default
             storage_override,
             n_distinct,
+            generated_expr,
+            identity,
         });
     }
     Ok(columns)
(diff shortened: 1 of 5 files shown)


NikolayS:
REV: feat/issue-54 — partitioned tables, types, domains, operators

CI: ✅ green (Build & Test passed)

---

**APPROVED — no blockers.**

---

## MINOR issues

### 1. Identity-sequence NO MAXVALUE heuristic is too broad — 

The NO MAXVALUE check:
```rust
let no_max = iseq.max_value.map(|m| m == i64::MAX || m >= 2_147_483_647).unwrap_or(true);
```
`2_147_483_647` is i32::MAX — a perfectly valid *explicit* MAXVALUE for a bigint identity column. A user who sets `MAXVALUE 2147483647` on a bigint sequence will get `NO MAXVALUE` in the dump, losing the intended cap. Suggest: only treat `i64::MAX` (and the small-integer sequence ceiling `9223372036854775807`) as NO MAXVALUE defaults; otherwise emit the literal value.

### 2. Domain DDL: NOT NULL emitted before DEFAULT — 

```rust
if d.not_null { out.push_str(" NOT NULL"); }
if !d.default_expr.is_empty() { out.push_str(&format!(" DEFAULT {}", ...)); }
```
PostgreSQL CREATE DOMAIN syntax puts DEFAULT before the NOT NULL/CHECK constraints (matching real pg_dump). The generated order (`NOT NULL DEFAULT ...`) is technically accepted by the parser, but it diverges from pg_dump output. Swap the two blocks.

### 3. Range type canonical/subtype_diff functions lack schema qualification — 

The subquery:
```sql
pf.proname || '(' || pg_catalog.format_type(pf.proargtypes[0], NULL) || ')'
```
emits `funcname(argtype)` with no schema prefix. For user-defined canonical or subtype_diff functions in non-`pg_catalog` schemas this would produce an unresolvable name on restore. Should join `pg_namespace` on `pf.pronamespace` and emit `schema.funcname`. (No current test exercises this path, so no CI failure, but it's a latent bug for custom range types.)

### 4. Test: `create_type_range` doesn't verify collation — 

The test setup creates `textrange AS RANGE (subtype = text, collation = "C")` but the test only checks for `subtype = text`. Add an assertion for `collation` to guard against the collation being silently dropped in future refactors.

### 5. Operator class `write_create_operator_class` always emits `STORAGE` — 

Real pg_dump only emits the `STORAGE` clause when the storage type differs from the indexed type; omitting it is the common case. The current implementation always appends `STORAGE type_name` even when unnecessary. Not blocking because `create_operator_class` is still `#[ignore]`d, but worth fixing before that test is un-ignored.

---

## Notes
- Partitioned table DDL (PARTITION BY, partition children) ✅ correct — emits `PARTITION OF … FOR VALUES FROM` for children, `PARTITION BY` for parents.
- Generated columns (`GENERATED ALWAYS AS … STORED`) ✅ — correctly clears `default_expr` so DEFAULT is not double-emitted.
- Identity columns ✅ — fix commit `a4a34db` correctly restores NOT NULL in column def and moves ALTER TABLE ADD GENERATED to immediately after CREATE TABLE.
- Domain creation ✅ — constraint list queried separately via `contypid`, emitted correctly.
- Enum/composite/range types ✅ — schema filtering consistent with other catalog functions.
- Operator family DDL ✅ — simple and correct.
- Identity sequence exclusion from `get_sequences` ✅ — `deptype = 'i'` filter is the right approach.

NikolayS:
## Code Review — PR #62: feat: partitioned tables, types, domains, operators

CI is green ✅. The core approach is solid and well-structured. Found several issues ranging from correctness bugs to polish items.

---

### Findings

**FINDING:**
- severity: HIGH
- confidence: 9
- file: src/dump/format.rs
- line: ~1013-1026 (write_alter_table_add_identity)
- issue: **Wrong NO MINVALUE / NO MAXVALUE detection for non-default sequence configurations.** The condition `m >= 2_147_483_647` will incorrectly emit `NO MAXVALUE` for any bigint identity sequence with a custom max between 2,147,483,647 and i64::MAX-1 (e.g. `MAXVALUE 5000000000` becomes `NO MAXVALUE`). For descending sequences (seqmin = -9223372036854775808, seqmax = -1), both checks fail: `no_min` is false (−big ≠ 1) and `no_max` is false (−1 < 2147483647), so it emits explicit `MINVALUE`/`MAXVALUE` instead of `NO MINVALUE`/`NO MAXVALUE` — diverging from real pg_dump.
- fix: Proper detection requires knowing the data type. For `bigint`: default min = 1 (asc) or -9223372036854775808 (desc); default max = 9223372036854775807 (asc) or -1 (desc). The most correct approach: store `seqincrement` and check if min/max equal the type-appropriate defaults based on sign of increment. Or follow pg_dump's approach: compare to `pg_sequence_parameters(seqrelid)` defaults. Minimally: fix the max check to `m == i64::MAX` only (removing the `>= 2_147_483_647` branch that catches non-default values).

---

**FINDING:**
- severity: HIGH
- confidence: 8
- file: src/dump/catalog.rs / src/dump/mod.rs
- line: ~722-736 (catalog), ~398-407 (mod.rs)
- issue: **Identity sequences silently skipped when `--table` filter is active.** `write_alter_table_add_identity` is only called when `!table_filter_active`. When a user runs `pg_plumbing -t test_table_identity`, the identity sequence definition is omitted. Real `pg_dump -t table` DOES emit the `ALTER TABLE ... ADD GENERATED ... AS IDENTITY` block.
- fix: Apply the same table-filter logic used for sequences: check `iseq.table_schema == table.schema && iseq.table_name == table.name` (which is already done) but remove the outer `!table_filter_active` guard, since we're already inside the per-table loop.

---

**FINDING:**
- severity: MEDIUM
- confidence: 8
- file: src/dump/catalog.rs
- line: ~693 (get_identity_sequences, seq_name construction)
- issue: **`seq_name` is built as raw string concatenation `sn.nspname || '.' || s.relname` without identifier quoting.** If the sequence or schema name contains uppercase letters, spaces, or reserved words, the `SEQUENCE NAME` clause in `ALTER TABLE ... ADD GENERATED ... AS IDENTITY` will be syntactically invalid or point to the wrong object.
- fix: Either use `quote_ident(sn.nspname) || '.' || quote_ident(s.relname)` in SQL, or apply quoting in `write_alter_table_add_identity` by splitting on `.` and calling `quote_ident` on each part. The same pattern already applies to `multirange_name` in `get_range_types`.

---

**FINDING:**
- severity: MEDIUM
- confidence: 7
- file: src/dump/catalog.rs
- line: ~882 (get_domains, constraint query ORDER BY)
- issue: **Domain constraints ordered by `conname` (alphabetically) instead of `oid` (creation order).** Real `pg_dump` preserves creation order by ordering on `c.oid`. For domains with multiple constraints, alphabetical ordering could produce different output than pg_dump — a constraint named `z_check` would appear after `a_check` regardless of definition order.
- fix: Change `ORDER BY conname` to `ORDER BY c.oid`.

---

**FINDING:**
- severity: MEDIUM
- confidence: 7
- file: src/dump/catalog.rs
- line: ~858 (get_range_types, canonical_func / subtype_diff_func SQL)
- issue: **Range type canonical/subtype_diff functions are constructed with only the first argument type and no schema qualification.** The SQL builds `proname || '(' || format_type(proargtypes[0], NULL) || ')'`. If the canonical function is in a user schema, the output won't be schema-qualified. Real pg_dump emits `schema.funcname` (schema-qualified name, no signature). Functions with overloads could also collide.
- fix: Use `pn.nspname || '.' || pf.proname` (JOIN on pg_namespace) instead of the signature-based form, matching pg_dump behavior. Only omit schema for `pg_catalog` functions.

---

**FINDING:**
- severity: MEDIUM
- confidence: 9
- file: src/dump/format.rs
- line: ~1551-1572 (write_create_operator_class)
- issue: **Generated `CREATE OPERATOR CLASS` SQL is invalid.** The body is `AS\n    STORAGE type;` — a STORAGE-only body with no OPERATOR or FUNCTION members. PostgreSQL requires at least one member for most access methods; STORAGE alone is only valid for GiST/GIN, not btree. This code is dead (the test remains `#[ignore]`) but would produce invalid SQL if ever reached.
- fix: Implement real operator class member emission (OPERATOR N, FUNCTION N entries) or leave as `#[ignore]` with a clear TODO comment instead of emitting broken SQL.

---

**FINDING:**
- severity: LOW
- confidence: 9
- file: src/dump/catalog.rs
- line: ~724-736 (get_identity_sequences, IdentitySequenceInfo struct)
- issue: **`min_value`/`max_value` fields are documented as `Option<i64>` with `None` meaning "NO MINVALUE/NO MAXVALUE", but the query always sets `Some(value)`.** The `Option` wrapper is misleading — the `None` path is never taken.
- fix: Either always use `i64` (non-optional) since you always have a concrete value from `pg_sequence`, or implement the proper detection described in the HIGH finding above and actually populate `None` when the value equals the type default.

---

**FINDING:**
- severity: LOW
- confidence: 8
- file: src/dump/catalog.rs
- line: ~860 (get_range_types, multirange_name)
- issue: **`multirange_name` is built as `mn.nspname || '.' || mt.typname` without identifier quoting**, same issue as `seq_name`. If schema/type name needs quoting, `multirange_type_name = schema.typename` will be syntactically wrong.
- fix: Use `quote_ident(mn.nspname) || '.' || quote_ident(mt.typname)` in SQL.

---

**FINDING:**
- severity: LOW
- confidence: 6
- file: src/dump/catalog.rs
- line: ~852 (get_composite_types), ~880 (get_domains)
- issue: **N+1 query pattern** — for each composite type a separate query fetches its fields; for each domain a separate query fetches its constraints. On databases with many types/domains this could be slow.
- fix: Not a correctness issue. Could be addressed later with a single join query using `array_agg`, similar to how `get_enum_types` aggregates labels in one query.

---

### VERDICT: CHANGES NEEDED

The PR is a significant and well-organized step forward — 13 real tests unlocked with proper setup, correct catalog queries for ENUMs/RANGEs/composite types/domains/opfamilies, and the identity sequence exclusion from regular sequences is correctly implemented.

However there are two HIGH severity correctness bugs:
1. **NO MINVALUE/NO MAXVALUE heuristic is wrong** for descending sequences and for large custom bigint max values — will produce incorrect SQL for those cases.
2. **Identity sequences silently dropped when `--table` filter is active** — the `!table_filter_active` guard should be removed since the inner loop already filters by table name.

The MEDIUM findings (unquoted identifiers in seq_name/multirange_name, constraint ordering, canonical function formatting) are output-correctness issues that diverge from real pg_dump.

Recommending changes before merge.

NikolayS:
## Code Review: PR #62 — feat: partitioned tables, types, domains, operators

Overall solid implementation covering a lot of ground. CI is green, tests pass. Found several issues ranging from a correctness bug that breaks restore fidelity, to style/edge-case concerns. Details below.

---

### FINDING 1

- **severity:** HIGH
- **confidence:** 9
- **file:** `src/dump/catalog.rs`
- **line:** 703–732 (`get_columns`)
- **issue:** Identity columns (`attidentity = 'a'` or `'d'`) have a `pg_attrdef` row whose expression is `nextval('seq'::regclass)`. The code only suppresses `default_expr` when `attgenerated == "s"` (stored generated columns). For identity columns, `default_expr` is left as `Some("nextval(...)")`, so `write_create_table` will emit `DEFAULT nextval(...)` inside `CREATE TABLE`. Then `write_alter_table_add_identity` emits `ALTER TABLE … ADD GENERATED … AS IDENTITY`. PostgreSQL rejects `ADD GENERATED` on a column that already has a non-identity `DEFAULT`, so the dump is **non-restorable** for identity columns.
- **fix:** Also clear `default_expr` when `attidentity != ""`:
  ```rust
  let default_expr: Option<String> = if attgenerated == "s" || !attidentity.is_empty() {
      None
  } else {
      row.get("default_expr")
  };
  ```

---

### FINDING 2

- **severity:** HIGH
- **confidence:** 8
- **file:** `src/dump/format.rs`
- **line:** 1636 (`write_alter_table_add_identity`)
- **issue:** `SEQUENCE NAME` is written using the unquoted SQL string `iseq.seq_name`, which is built in the catalog query as `sn.nspname || '.' || s.relname` — bare string concatenation with no `quote_ident`. If either the schema or sequence name contains uppercase letters, spaces, or special characters, the emitted `SEQUENCE NAME` will be syntactically invalid.
- **fix:** Build `seq_name` in Rust using `quote_ident` (or do it in the format function):
  ```rust
  // In get_identity_sequences, store schema + name separately, then:
  let seq_q = format!("{}.{}", quote_ident(&iseq.seq_schema), quote_ident(&iseq.seq_relname));
  out.push_str(&format!("    SEQUENCE NAME {seq_q}\n"));
  ```

---

### FINDING 3

- **severity:** MEDIUM
- **confidence:** 9
- **file:** `src/dump/format.rs`
- **line:** 1643–1646 (`write_alter_table_add_identity`)
- **issue:** The NO MINVALUE / NO MAXVALUE heuristic is wrong in two ways:
  1. **Descending sequences**: default `min_value` is `i64::MIN` (not `1`), so the check `min_value == 1` is `false` and it emits `MINVALUE -9223372036854775808` instead of `NO MINVALUE`.
  2. **Smallint identity**: default `max_value` is `32767`; the check `m >= 2_147_483_647` is `false`, so it emits `MAXVALUE 32767` instead of `NO MAXVALUE`.
  3. **Custom values in (2^31-1, 2^63-1)**: any explicit `MAXVALUE` in that range is incorrectly suppressed as `NO MAXVALUE`.
- **fix:** The reliable approach (matching real `pg_dump`) is to compare against the sequence type's actual type default bounds. Alternatively, store `increment_by` and check: for ascending (`increment_by > 0`): `no_min = (min == 1)`, `no_max = (max == type_max_for_type)`; for descending: `no_min = (min == type_min)`, `no_max = (max == -1)`. The sequence's data type can be obtained from `pg_sequence.seqtypid`.

---

### FINDING 4

- **severity:** MEDIUM
- **confidence:** 8
- **file:** `src/dump/catalog.rs`
- **line:** ~2990–2995 (range type query, `canonical_func` / `subtype_diff_func` sub-selects)
- **issue:** The canonical and subtype_diff functions are formatted as `pf.proname || '(' || ... || ')'` — only the bare function name, **not schema-qualified**. If those functions live outside `pg_catalog` and outside the current `search_path`, the emitted `canonical = myfunc(bigint)` would fail to resolve at restore time.
- **fix:** Join `pg_namespace` on `pf.pronamespace` and prepend the schema:
  ```sql
  COALESCE((
      SELECT pn.nspname || '.' || pf.proname || '(' || ... || ')'
      FROM pg_catalog.pg_proc pf
      JOIN pg_catalog.pg_namespace pn ON pn.oid = pf.pronamespace
      WHERE pf.oid = rg.rngcanonical
  ), '') AS canonical_func
  ```

---

### FINDING 5

- **severity:** MEDIUM
- **confidence:** 7
- **file:** `src/dump/catalog.rs`
- **line:** ~2995 (range type query, `multirange_name`)
- **issue:** `multirange_name` is built in SQL as `mn.nspname || '.' || mt.typname` — unquoted. In `write_create_range_type` it is interpolated directly into the DDL. For multirange types in schemas with special characters, this produces invalid SQL.
- **fix:** Build `multirange_name` using `quote_ident` in Rust after fetching schema and name as separate columns, or add quoting in SQL via `quote_ident()` (PG function).

---

### FINDING 6

- **severity:** LOW
- **confidence:** 9
- **file:** `src/dump/catalog.rs`
- **line:** 59–64 (struct `ColumnInfo`)
- **issue:** `ColumnInfo.identity: Option<char>` is populated in `get_columns` but **never read** anywhere in `format.rs` or `mod.rs`. The identity column rendering is handled entirely through `IdentitySequenceInfo`. This is dead code that adds confusion about what drives identity column output.
- **fix:** Remove the `identity` field from `ColumnInfo` entirely, or add a `#[allow(dead_code)]` with a comment explaining it is reserved for future use (e.g., emitting `GENERATED ALWAYS AS IDENTITY` inline in `CREATE TABLE` without the separate ALTER path).

---

### FINDING 7

- **severity:** LOW
- **confidence:** 8
- **file:** `src/dump/format.rs`
- **line:** 1602–1607 (`write_create_operator_class`)
- **issue:** The emitted `CREATE OPERATOR CLASS` statement contains only `STORAGE {type_name}` as its body — no `OPERATOR` or `FUNCTION` members. This is syntactically valid but semantically wrong/incomplete for any real operator class. When the corresponding `create_operator_class` test is eventually un-ignored, this will produce a non-functional operator class.
- **fix:** This appears to be a known stub (test remains `#[ignore]`). The comment is clear. Just ensure the TODO is tracked in an issue before the ignore is removed.

---

### FINDING 8

- **severity:** LOW
- **confidence:** 8
- **file:** `src/dump/catalog.rs`
- **line:** 3371–3381 (`get_identity_sequences`)
- **issue:** `min_value` and `max_value` in `IdentitySequenceInfo` are typed `Option<i64>` but are always constructed as `Some(...)`. The `Option` wrapping is misleading (never `None` in practice) and causes `unwrap()` calls in `format.rs` (lines 1650, 1655).
- **fix:** Change the fields to plain `i64`, remove the `Option` wrapper, and eliminate the `unwrap()` calls. Or document explicitly why the `Option` is kept (reserved for `NO MINVALUE`/`NO MAXVALUE` sentinel meaning).

---

### FINDING 9

- **severity:** LOW
- **confidence:** 7
- **file:** `tests/pg_dump/t002_pg_dump.rs`
- **line:** Various un-ignored tests (e.g. `create_table_identity`, lines 1738–1749)
- **issue:** Tests check only for string presence in the dump output (e.g. `stdout.contains("GENERATED ALWAYS AS IDENTITY")`). They do **not** verify restore correctness — a dump with `DEFAULT nextval(...)` AND `ADD GENERATED AS IDENTITY` would still pass all these tests while being non-restorable.
- **fix:** Add a restore-and-verify step (or at minimum assert that the output does **not** contain `DEFAULT nextval` for identity columns). A smoke-test that pipes the dump into `psql` in a fresh database would catch this class of bugs systematically.

---

### FINDING 10

- **severity:** LOW
- **confidence:** 6
- **file:** `tests/common/mod.rs`
- **line:** ~865 (`setup_issue54_schema`)
- **issue:** The `textrange` RANGE type is created with `collation = "C"`. If the test instance doesn't support range type collation on `text` (which requires the `C` collation to be present), this step will fail silently only if the `psql` step's exit code is non-zero. The `"C"` collation is standard, so this is low risk — but the test setup uses `CASCADE` drops on types that may not exist yet (which is correct), while the operator family drop `IF EXISTS` is done before its creation (also correct).
- **fix:** No immediate action needed; low risk. Consider adding `\if` checks or separating the `DROP`/`CREATE` steps if portability across PG versions becomes a concern.

---

## Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | **HIGH** | Identity columns emit `DEFAULT nextval()` → non-restorable dump |
| 2 | **HIGH** | `SEQUENCE NAME` uses unquoted SQL string concatenation |
| 3 | MEDIUM | NO MINVALUE/MAXVALUE heuristic wrong for descending seqs & smallint |
| 4 | MEDIUM | Range type canonical/subtype_diff functions not schema-qualified |
| 5 | MEDIUM | `multirange_name` built without quoting in SQL |
| 6 | LOW | `ColumnInfo.identity` field is dead code |
| 7 | LOW | `CREATE OPERATOR CLASS` emits incomplete stub DDL (acknowledged) |
| 8 | LOW | `Option<i64>` fields always `Some(...)` → misleading + `unwrap()` |
| 9 | LOW | Tests check string presence only; restore correctness not validated |
| 10| LOW | Test setup minor portability notes |

## VERDICT: CHANGES NEEDED

Finding #1 (identity col `DEFAULT nextval` in CREATE TABLE) is a correctness bug that would make the dump non-restorable for any table with `GENERATED AS IDENTITY` columns. Finding #2 (unquoted sequence name) similarly breaks restore for non-standard identifiers. Both should be fixed before merging. The remaining issues are lower priority but worth addressing.

NikolayS:
Closing: the changes from this branch were already merged into main via PR #61 (squash merge). The  commit is also already in main. No unique code remains.

PR #65, opened 2026-04-04
fix: address review feedback on PR #62 (identity DEFAULT leak, quoting, heuristic)

NikolayS:
## Goal
Addresses 5 review comments raised on PR #62 (partitioned tables/types/domains/operators).

### Fix 1 (HIGH – blocking): Identity column DEFAULT leak
- **File:** `src/dump/catalog.rs` → `get_columns()`
- Identity columns store a `nextval(...)` expression in `pg_attrdef`.  The old code did **not** clear `default_expr` for identity columns, causing `CREATE TABLE` to emit `DEFAULT nextval(...)` **and** `ALTER TABLE … ADD GENERATED … AS IDENTITY` — which PostgreSQL rejects with a duplicate default error.
- Fix: mirror the existing generated-column guard: set `default_expr = None` when `attidentity != ""`.

### Fix 2 (HIGH – blocking): Unquoted identity sequence name
- **File:** `src/dump/catalog.rs` → `get_identity_sequences()` SQL
- `seq_name` was built as `sn.nspname || '.' || s.relname` — bare concatenation, breaks for uppercase/special-char schemas or sequences.
- Fix: `quote_ident(sn.nspname) || '.' || quote_ident(s.relname)`.

### Fix 3 (MEDIUM): NO MINVALUE / NO MAXVALUE heuristic wrong
- **File:** `src/dump/format.rs` → `write_alter_table_add_identity()`
- Old heuristic `m >= 2_147_483_647` failed for descending sequences and bigint sequences with a custom max between 2.1B and `i64::MAX-1`.
- Fix: match pg_dump's logic — compare against the type's actual bounds (int2/int4/int8) and check ascending vs descending direction.  Added `seqtype` field to `IdentitySequenceInfo` (fetched via `seq.seqtypid::regtype::text`).

### Fix 4 (MEDIUM): Range type functions not schema-qualified
- **File:** `src/dump/catalog.rs` → `get_range_types()` SQL
- `canonical_func` and `subtype_diff_func` were built with only `pf.proname` — no schema join.  Breaks for user-defined canonical/diff functions outside `pg_catalog`.
- Fix: join `pg_namespace` and emit `quote_ident(schema) || '.' || quote_ident(funcname)`.

### Fix 5 (MEDIUM): multirange_name unquoted
- **File:** `src/dump/catalog.rs` → `get_range_types()` SQL
- `multirange_name` was `mn.nspname || '.' || mt.typname` without quoting.
- Fix: `quote_ident(mn.nspname) || '.' || quote_ident(mt.typname)`.

## How to verify
```
cargo build   # must succeed, no warnings
cargo test    # 28 unit tests + 7 integration tests, all OK
```

Create a table with an identity column and verify the dump does **not** contain both `DEFAULT nextval(...)` and `ADD GENERATED … AS IDENTITY` for the same column.

### DIFF
--- a/src/dump/catalog.rs
+++ b/src/dump/catalog.rs
@@ -711,8 +711,11 @@ async fn get_columns(client: &Client, table_oid: u32) -> Result<Vec<ColumnInfo>>
             None
         };
 
-        // For generated columns, clear default_expr so we don't also emit DEFAULT.
-        let default_expr: Option<String> = if attgenerated == "s" {
+        // For generated columns and identity columns, clear default_expr so we
+        // don't also emit DEFAULT.  Identity columns store a `nextval(...)` in
+        // pg_attrdef which must NOT appear in CREATE TABLE — only the
+        // `ALTER TABLE … ADD GENERATED … AS IDENTITY` statement that follows.
+        let default_expr: Option<String> = if attgenerated == "s" || !attidentity.is_empty() {
             None
         } else {
             row.get("default_expr")
@@ -2916,6 +2919,8 @@ pub struct IdentitySequenceInfo {
     pub cache_size: i64,
     /// Whether the sequence cycles.
     pub cycle: bool,
+    /// Sequence data type: 'b' = bigint, 'i' = int4, 's' = int2.
+    pub seqtype: char,
 }
 
 /// Query ENUM types from the catalog.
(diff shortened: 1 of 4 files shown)


NikolayS:
## Code Review — PR #65 (fix: address review feedback on PR #62)

**Reviewer:** automated senior review
**CI status:** no checks reported (branch has no CI run yet)

---

### Fix-by-fix analysis

#### Fix 1 (HIGH) — Identity column DEFAULT leak ✅ CORRECT

`catalog.rs` → `get_columns()` now sets `default_expr = None` when `attidentity != ''` (or `attgenerated == 's'`). Logic is clean and correct. Also correctly excludes identity sequences from `get_sequences()` via `deptype='i'` filter.

#### Fix 2 (HIGH) — `seq_name` built without `quote_ident` ✅ CORRECT

SQL now uses `quote_ident(sn.nspname) || '.' || quote_ident(s.relname)`. Correct.

#### Fix 3 (MEDIUM) — NO MINVALUE / NO MAXVALUE heuristic ⚠️ STILL WRONG

**FINDING**
- **severity:** MEDIUM
- **confidence:** 9
- **file:** `src/dump/format.rs` → `write_alter_table_add_identity()`
- **issue:** The `no_min` check has the ascending and descending defaults swapped. The actual pg_dump logic (verified against `src/bin/pg_dump/pg_dump.c` REL_16_STABLE):

  | direction  | default_min | default_max |
  |-----------|-------------|-------------|
  | ascending  | **1**       | type_max    |
  | descending | **type_min**| **-1**      |

  The PR has:
  ```rust
  let no_min = iseq.min_value.map(|m| if ascending { m == type_min } else { m == 1 })
  ```
  This is **backwards for min**: ascending should check `m == 1`, descending should check `m == type_min`.

  The `no_max` side (`ascending → type_max`, `descending → -1`) is correct.

- **impact:** Every standard ascending identity column (the common case, seqmin=1, bigint) will emit `MINVALUE 1` instead of `NO MINVALUE`, diverging from pg_dump output. Descending sequences will emit `MINVALUE -9223372036854775808` instead of `NO MINVALUE`.

- **fix:**
  ```rust
  let no_min = iseq.min_value.map(|m| if ascending { m == 1 } else { m == type_min }).unwrap_or(true);
  ```

#### Fix 4 (MEDIUM) — Range type canonical/subtype_diff not schema-qualified ✅ CORRECT

Both `canonical_func` and `subtype_diff_func` are now built with a `pg_namespace` join and `quote_ident(pn.nspname) || '.' || quote_ident(pf.proname)`. Correct.

#### Fix 5 (MEDIUM) — `multirange_name` not quoted ✅ CORRECT

SQL now uses `quote_ident(mn.nspname) || '.' || quote_ident(mt.typname)`. Correct.

---

### New issues introduced

**FINDING**
- **severity:** LOW
- **confidence:** 7
- **file:** `src/dump/mod.rs` (identity sequence emit loop)
- **issue:** The identity sequence `ALTER TABLE … ADD GENERATED` is guarded by `if !table_filter_active`. When a user uses `-t` (table filter), identity column definitions are silently dropped from the dump. This means `pg_dump -t my_identity_table` produces a restore script that creates the column as plain integer without the identity property.
- **note:** This is pre-existing behavior from PR #62, not introduced by #65. Worth a follow-up issue but not a blocker for this PR.

---

### VERDICT: CHANGES NEEDED

Fix 3 (`no_min` heuristic) is still incorrect. The ascending/descending min defaults are swapped relative to pg_dump's actual logic. For the common case of a standard ascending bigint identity column, this emits `MINVALUE 1` instead of `NO MINVALUE`.

Fixes 1, 2, 4, 5 are all correct and complete.

The one-line fix:
```rust
// In write_alter_table_add_identity(), replace:
let no_min = iseq.min_value.map(|m| if ascending { m == type_min } else { m == 1 }).unwrap_or(true);
// With:
let no_min = iseq.min_value.map(|m| if ascending { m == 1 } else { m == type_min }).unwrap_or(true);
```

NikolayS:
Fixed the remaining review blocker: identity sequence MINVALUE logic now matches pg_dump (ascending default min = 1, descending default min = type_min). Added coverage in create_table_identity to assert NO MINVALUE.\n\nVerification:\n- cargo fmt -- --check\n- cargo test create_table_identity -- --nocapture\n- cargo test\n- GitHub CI Build & Test: green

